PR #6, opened 2020-04-27
build: Fix broken `install` and `dist` targets

xelxebar:
Previously the automake variable `dist_pkgdata_DATA` was set to a list
of non-existing `*.mm` databases. This variable controls which files are
installed to `$(datadir)/metamath` during the make targets `install` and
`dist`, which were failing with an error due to the missing
dependencies.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -36,14 +36,6 @@ metamath_SOURCES = \
 	mmwtex.c \
 	$(noinst_HEADERS)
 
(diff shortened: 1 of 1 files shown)


nmegill:
@david-a-wheeler Does this seem reasonable to you?

david-a-wheeler:
Yes it does! Metamath the program should be distinct from the data it manages.

Issue #8, opened 2021-12-24
Metamath software future directions


wlammen:
Currently the development of metamath.exe is orphanized.  The discussion in pull request https://github.com/metamath/set.mm/pull/2392 showed, there is interest in moving forward, and there exist already some ideas under the hood.

This issue is meant to evaluate the current state of affairs, and bundle further discussions in a single location, instead of hijacking other threads for this purpose.

To me, there seems to be neither a team dedicated to, nor matured results about, future directions of Metamath.  Lets change this starting a discussion from scratch.  As far as I am concerned, I will be busy over Xmas, but will have spare time afterwards.

Please post your ideas, regardless of whether they are blurred or laser-sharp.  I am looking forward to learn about your visions.

Ideally, I should have started with my own ideas first, but I am in a bit of a hurry right now. Sorry.

Wolf

wlammen:
If you don't like the idea of an inner circle, and want a broader public to share thoughts, consider opening a discussion in Google groups.

tirix:
Thanks for starting this thread!

- I'd like to share here a link to [`metamath-knife`](https://github.com/david-a-wheeler/metamath-knife), which ambition it is to become a widely usable library for Metamath, with functions similar to `metamath-exe`.
- Before any major refactoring is done, I would like to get the opportunity to merge here changes I have done to generate the ["structured typesetting" pages](http://metamath.tirix.org/effsumlt.html). At the time I did these changes, Norm had not yet setup this repository, and he asked me to hold on my changes until he was ready. I'll try to open a PR shortly.

savask:
I think everyone agrees that a general cleanup of the codebase is necessary. I would suggest:

- Removing obsolete features, for example, mmmaci.c (Macintosh interface functions, whatever that is)
- Removing obsolete code - metamath-exe files often store older revisions in comments, for example, here are a few lines from mmpars.c:

```
char *readRawSource(/*vstring inputFn,*/ /* 2-Feb-2018 nm Unused argument */
         vstring fileBuf, /* added 31-Dec-2017 */
         /*long bufOffsetSoFar,*/  /* deleted 31-Dec-2017 nm */
         long *size /* 31-Dec-2017 Now an input argument */)
{
  long charCount = 0;
  /*char *fileBuf;*/ /* 31-Dec-2017 */
  char *fbPtr;
  /*long lineNum;*/ /* 2-Feb-2018 Now computed in rawSourceError() */
  flag insideComment;
[...]

PR #9, opened 2021-12-24
Structured Typesetting (STS) generation

tirix:
I've tried to make as few changes as possible. The changes are in the following source files:
* `metamath.c` for the changes to the `SHOW STATEMENT` command and the new `VERIFY STS` command, as well as the output post-processing,
* `mmcmdl.c` for the new `HELP` command options, and the new commands,
* `mmcmds.c` for the changes to the `SHOW STATEMENT` command at top level (distinct and dummy variables, syntax hints),
* `mmdata.c` for some utility functions,
* `mmhlpa.c` and `mmhlpb.c` for the new built-in `HELP` options,
* **`mmhtbl.c`**, a new file with a hash table implementation,
* **`mmwsts.c`, a new file with the main STS implementation**,
* `mmwtex.c`, for the main hook into the STS formula output and for in-line comment formulas.

This also adds a `.gitignore` file to ignore object files.


wlammen (commented):
Why did you add the makefile? The readme instructions note several ways to compile metamath, including gcc m*.c -o metamath, and using automake.

wlammen (commented):
metamath.c line 3032 duplicated code from line 2393

wlammen (commented):
style, readability command line option STS: Norm used always verbose forms, consider expanding it to something like STRUCTURED_TYPESETTING or so

wlammen (commented):
built-in help does not cover new option

tirix:
Thank you Wolf for your review!

> built-in help does not cover new option

It actually does, see changes in `mmhlpa.c` and `mmhlpb.c`.

> style, readability command line option STS: Norm used always verbose forms, consider expanding it to something like STRUCTURED_TYPESETTING or so

That's very verbose, though. Maybe just "STRUCTURED" ?

> Why did you add the makefile? The readme instructions note several ways to compile metamath, including gcc m*.c -o metamath, and using automake.

Sure, I can remove the makefile.

benjub:
How do you render this "structured typesetting"? Is this MathML, MathJax or something like this ?

For texts, and in particular help texts, Norm used the double-space convention between sentences, and I admit I like that.

tirix:
> How do you render this "structured typesetting"? Is this MathML, MathJax or something like this ?

Yes, the [`set-mathml.mmts`](https://github.com/metamath/set.mm/blob/develop/set-mathml.mmts) file contains instructions about how to display set.mm formulas as MathML. Then, the MathML result is sent to MathJax for rendering.
The actual MathJax command used is included in that file in the `$c ... $.` instruction towards the end of the file. That instruction is read and executed by the metamath-exe program, with the input file generated send it its standard input.

> For texts, and in particular help texts, Norm used the double-space convention between sentences, and I admit I like that.

Your wish has come true with the last commit!

mmhlpb.c
```diff
@@ -710,6 +718,16 @@ H("see the 21-Dec-2017 entry in http://us.metamath.org/mpeuni/mmnotes.txt .");
 H("");
 
 
+g_printHelp = !strcmp(saveHelpCmd, "HELP VERIFY STS");
+H("Syntax:  VERIFY STS <format>");
+H("");
+H("This command error-checks that the STS rules definition covers all syntax");
+H("defined in the Metamath source file loaded. It runs through all non");
```
benjub, line 725: Double space here !  And I would write "non-definitional" with a hyphen, or even "nondefinitional", since "non" is not a word.

benjub:
Thanks. So maybe you can use the option "/ MATHML" ?  I'm fine with "/ STRUCTURED" too.  I cannot approve this PR since I am not fluent in C :-(

wlammen:
The option STS or /STRUCTURED is the second alternative to /HTML and /ALT_HTML.  Are you able to roughly tell in what way these options differ by just looking at the name?  If not, people will have to guess at what the result of invocations are.  Do you expect this option to be typed by users of metamath directly?  Or will it be likely be part of a script rarely changing?  If you type the option frequently, a short name is handy, otherwise even a very long option name will hardly annoy anyone, but is easily understandable to casual readers of the script.
If I understand your code right, embedded snippets are translated via a sort of grammar file into true HTML.  This process is then best expressed in your option tag.

I am still busy with Xmas, have just skimmed your PR.  I currently cannot look into this further.

tirix:
> Thanks. So maybe you can use the option "/ MATHML" ? I'm fine with "/ STRUCTURED" too.

Actually I wanted to keep it generic because one could generate anything with it, not just MathML.
The `set-mathml.mmts` just happens to contain rules to generate MathML, the Metamath.exe functionality is completely unaware of what it is (actually, I had written another file to generate LaTeX with the same method).

tirix:
> The option STS or /STRUCTURED is the second alternative to /HTML and /ALT_HTML. 

Yes, and it is itself followed by the name of the production to use. A command would typically be:
``` SHOW STATEMENT syl / STS mathml ```
which instructs the program to read and execute the `mathml` instruction file (`set-mathml.mmts`).

If you had, say, a `graphviz-dot` instruction file, the command:
``` SHOW STATEMENT syl / STS graphviz-dot ```
would search a `set-graphviz-dot.mmts` file, and presumably output some graphs instead of formulas (that might be an interesting experiment!).

> If I understand your code right, embedded snippets are translated via a sort of grammar file into true HTML. This process is then best expressed in your option tag.

Yes, that's roughly the process. So if I follow you, we could use e.g. /STS_HTML as an option tag?

**Comment by wlammen** [comment]
What about EXPAND_HTML?

**Comment by benjub** [comment]
When running metamath, it is enough to type any non-ambiguous prefix, so here, `MM> show statement syl /S mathml` would suffice. In other words, there is no big inconvenience in having a verbose option. As for option names, I always thought the other two (HTML, ALT_HTML) were not optimal.  I think more explicit choices would be /UNICODE and /GIF.

**Review by wlammen — COMMENTED** [review]
.gitignore: OK, see here (https://stackoverflow.com/questions/6626136/best-practice-for-adding-gitignore-to-repo)
Maybe one should put this info into the README?

**Review by wlammen — CHANGES_REQUESTED** [review]
metamath.c line 60 suggest a version number and add a changelog entry.

**Review by wlammen — CHANGES_REQUESTED** [review]
naming consistency mmwtex.h line 33:  stsFlag (and possibly following) should be prefixed with g_ (g_stsFlag, matching g_altHtmlFlag for example), see changelog 0.187 metamath.c, line 88

**Review by wlammen — CHANGES_REQUESTED** [review]
optimization metamath.c line 2381: The function switchPos("/ STS") is called three times more or less in succession. Consider evaluating the result once and use a local variable to recall the result within the if conditions.

**Review by wlammen — COMMENTED** [review]
parsetSTSRules type: why parset...?  Looks like a typo

**Review by wlammen — CHANGES_REQUESTED** [review]
meamath.c line 2977:  Suspicious value of i.  The other options / TIME or / NO_VERSIONING add 2 to the number of options, seemingly counting the slash and the tag as different entries.  Why is this not done for / STS?  If the count is correct, IMO a comment should clarify the underlying logic.

to be continued...

**Comment by tirix** [comment]
@wlammen I'm going to copy your remarks directly into a code review, it creates a sub-thread per remark and I think it's much easier to follow like that.

File: metamath.c [hunk 2]
```diff
@@ -2959,6 +2973,8 @@ void command(int argc, char *argv[])
       i = 5;  /* # arguments with only / HTML or / ALT_HTML */
       if (noVersioning) i = i + 2;
       if (switchPos("/ TIME")) i = i + 2;
+      /* 7-Jul-2017 added MathML/STS */
+      if (switchPos("/ STS")) i = i + 1;
```
tirix: @wlammen wrote:

> Suspicious value of i. The other options / TIME or / NO_VERSIONING add 2 to the number of options, seemingly counting the slash and the tag as different entries. Why is this not done for / STS? If the count is correct, IMO a comment should clarify the underlying logic.

Interesting. So, a typical `SHOW STATEMENT` command looks like this: 
``` SHOW STATEMENT syl / HTML ```
That would be g_rawArgs = 5, which already includes 2 arguments for the `/ HTML`.
As you correctly guessed, for `/ NO_VERSIONING` and `/ TIME`, there are two more arguments, therefore the +2.
In the case of STS, the `/ STS` would already be accounted for in the 5 (taking the place of the `/ HTML`), so that's not why the +1 is for. Rather, in the case of STS, there is one more argument, namely the name of the output processing, e.g. `mathml`. That is what the +1 is for.

I'll add a comment to make that clearer.

**Comment by tirix** [comment]
> metamath.c line 60 suggest a version number and add a changelog entry

Ok, I've updated the history and proposed a version 0.199.
This kind of history entry can only really be finalized at the merge time (imagine another PR comes before) and typically will create merge conflicts, though.

File: .gitignore [hunk 3]
```diff
@@ -0,0 +1,2 @@
+*.o
+metamath
```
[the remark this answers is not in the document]
wlammen: The reason why I added this comment is, because the metamath sources are distributed via a tar archive as well. Obviously you wont need the gitignore in this case. I think, some kind of how-to is finally called for.  That are my thoughts.  I was curious why you added the file.  Separate PR is fine with me.

mmwtex.h
```diff
@@ -29,6 +29,10 @@ extern flag g_htmlFlag;  /* HTML flag: 0 = TeX, 1 = HTML */
 extern flag g_altHtmlFlag;  /* Use "althtmldef" instead of "htmldef".  This is
     intended to allow the generation of pages with the old Symbol font
     instead of the individual GIF files. */
+/* 19-Jul-2017 tar Added for STS/MathML output */
+extern flag stsFlag; /* STS output (for "structural typesetting") */
+extern vstring stsOutput; /* output mode chosen for STS (follows STS flag) */
+extern vstring postProcess; /* command to pipe the output into (used for MathJax prerendering) */
```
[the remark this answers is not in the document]
wlammen: This are the current rules.

metamath.c
```diff
@@ -2375,14 +2376,25 @@ void command(int argc, char *argv[])
       g_htmlFlag = 1;
       /* If not specified, for backwards compatibility in scripts
          leave g_altHtmlFlag at current value */
+      /* 7-Jul-17 - MathML/STS handling */
       if (switchPos("/ HTML") != 0) {
-        if (switchPos("/ ALT_HTML") != 0) {
-          print2("?Please specify only one of / HTML and / ALT_HTML.\n");
+        if (switchPos("/ ALT_HTML") != 0 || switchPos("/ STS") != 0 ) {
+          print2("?Please specify only one of / HTML , / ALT_HTML and / STS.\n");
```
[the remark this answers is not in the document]
wlammen: You're right.  It still looks awkward since the call and the parameter is coded several times.

metamath.c
wlammen: I am subject to a learning curve as well while reviewing, wrt both to your source code and the review style.

mmdata.c
```diff
@@ -1101,6 +1101,39 @@ int nmbrEq(nmbrString *s,nmbrString *t)
...
 
 
+/* Compute a hash of a string */
+/* This simply computes a XOR of the first numbers */
+int nmbrHash(nmbrString *s)
+{
+  static long salt[] = { 4938, 48977, 6897, 7293, 2663, 7925, 2999, 12238,
+      40033, 14038, 10699, 29746, 56108, 34526, 63576, 52053, 61949, 41177, 43740, 22822
+ };
+  long i;
```
wlammen, line 1111: Initialize with value from line 1113

mmdata.c
```diff
@@ -1101,6 +1101,39 @@ int nmbrEq(nmbrString *s,nmbrString *t)
 }
 
 
+/* Compute a hash of a string */
+/* This simply computes a XOR of the first numbers */
+int nmbrHash(nmbrString *s)
+{
+  static long salt[] = { 4938, 48977, 6897, 7293, 2663, 7925, 2999, 12238,
```
wlammen, line 1108: Only 13 (?) values actually needed.

wlammen (changes requested):

mmdata.c
```diff
@@ -1101,6 +1101,39 @@ int nmbrEq(nmbrString *s,nmbrString *t)
...
+/* Unlike strcmp, this returns a 1 if the strings are equal
+   and 0 otherwise. */
+/* Only the token is compared.  The whiteSpace string is
+   ignored. */
+int nmbrSubEq(nmbrString *s,long sstart, nmbrString *t, long tstart, long len)
+{
+  long i;
+  if (sstart - 1 + len > nmbrLen(s)
+   || tstart - 1 + len > nmbrLen(t)) return 0;
+  for (i = 0; s[sstart-1+i] == t[tstart-1+i] && i<len; i++);
```
wlammen, line 1132: Suspicious index in first loop: -1, if sstart == 0.  (1) Out of bounds access to s and t, or (2) Parameters sstart and tstart must be > 0.

Either provide parameter checks (cf. line 1211), or (minimum) state limitations in comment line 1122

mmdata.c
```diff
@@ -1170,6 +1203,34 @@ nmbrString *nmbrSpace(long n)
   return (sout);
 }
 
+/* Search for the nth occurrence of string2 in string1 */
+long nmbrInstrN(long start_position, long occ, nmbrString *string1,
+  nmbrString *string2, long start2, long length2)
+{
+  if (start_position < 1) start_position = 1;
```
wlammen, line 1210: This means: If garbage is provided in parameter start_position, then I sanitize it to something more useful, hoping for the best.  This kind of fault tolerance supports a sloppy programming on the caller's side.  Better throw an exception, or call bug()

mmdata.c
```diff
@@ -1170,6 +1203,34 @@ nmbrString *nmbrSpace(long n)
   return (sout);
 }
 
+/* Search for the nth occurrence of string2 in string1 */
```
wlammen, line 1206: Explain parameters (semantics and limitations) in a comment.  For example, what does occ mean? n would match the functions name.

mmdata.c
```diff
@@ -1170,6 +1203,34 @@ nmbrString *nmbrSpace(long n)
...
 
+/* Search for the nth occurrence of string2 in string1 */
+long nmbrInstrN(long start_position, long occ, nmbrString *string1,
+  nmbrString *string2, long start2, long length2)
+{
+  if (start_position < 1) start_position = 1;
+  start_position--;
+  for(; occ > 0;occ--) {
+    long ls1, i, j;
+    ls1 = nmbrLen(string1);
```
wlammen, line 1214: Pull constant evaluations out of the loop.

mmdata.c
```diff
@@ -1170,6 +1203,34 @@ nmbrString *nmbrSpace(long n)
...
 }
 
+/* Search for the nth occurrence of string2 in string1 */
+long nmbrInstrN(long start_position, long occ, nmbrString *string1,
+  nmbrString *string2, long start2, long length2)
+{
+  if (start_position < 1) start_position = 1;
+  start_position--;
+  for(; occ > 0;occ--) {
+    long ls1, i, j;
```
wlammen, line 1213: declare i and j in for commands

mmdata.c
```diff
@@ -1170,6 +1203,34 @@ nmbrString *nmbrSpace(long n)
...
+    for (i = start_position - 1; i <= ls1 - length2; i++) {
+      flag found = 1;
+      for (j = 0; j < length2; j++) {
+        if (string1[i+j] != string2[start2-1+j]) {
+          found = 0;
+          break;
+        }
+      }
+      if (found) {
+	start_position = i+1;
```
wlammen, line 1225: incorrect indentation

mmdata.c
```diff
@@ -1170,6 +1203,34 @@ nmbrString *nmbrSpace(long n)
...
+  if (start_position < 1) start_position = 1;
+  start_position--;
+  for(; occ > 0;occ--) {
+    long ls1, i, j;
+    ls1 = nmbrLen(string1);
+    if(start_position++ >= ls1) return 0;
+    for (i = start_position - 1; i <= ls1 - length2; i++) {
+      flag found = 1;
+      for (j = 0; j < length2; j++) {
+        if (string1[i+j] != string2[start2-1+j]) {
```
wlammen, line 1219: This condition is usually part of the for-command.

mmdata.c
```diff
@@ -1548,6 +1609,21 @@ nmbrString *nmbrAddElement(nmbrString *g, long element)
 }
 
 
+/* Add a single number to start of a nmbrString - faster than nmbrCat */
+nmbrString *nmbrUnshiftElement(nmbrString *g, long element)
+{
+  long length;
```
wlammen, line 1615: declaration should be initialization

mmdata.c
```diff
@@ -1548,6 +1609,21 @@ nmbrString *nmbrAddElement(nmbrString *g, long element)
 }
 
 
+/* Add a single number to start of a nmbrString - faster than nmbrCat */
```
wlammen, line 1612: The return value is pushed on an internal variable stack with an implicit memory management.  This important detail is not mentioned in the comment.

mmdata.h
```diff
@@ -281,6 +281,10 @@ long nmbrLen(nmbrString *s);
 long nmbrAllocLen(nmbrString *s);
 void nmbrZapLen(nmbrString *s, long length);
 
+/* Search for the nth occurrence of string2 in string1 */
```
wlammen, line 284: Should explain parameters and specify their limitations.

wlammen (changes requested):

mmhtbl.c
```diff
@@ -0,0 +1,184 @@
...
+  unsigned int bucket = hashtable->hashFunc(key) % hashtable->bucket_count;
+
+  /* Look for the key */
+  int *pli = &hashtable->buckets[bucket];
+  while(*pli != NO_LINKEDITEM) {
+    if(!hashtable->eqFunc(linkedItems[*pli].key, key)) {
+      /* Found it, free the object and remove it from the chain */
+      hashtable->freeFunc(&linkedItems[*pli].key, &linkedItems[*pli].object);
+      int old_free = free_linkedItem;
+      free_linkedItem = *pli;
```
wlammen, line 145: DRY duplicated code from line 76

wlammen (changes requested):

tirix:
@wlammen thank you very much for your careful review! 
I think you are not quite done yet: please give me a heads up when you are, and I will try to address all remarks at once.

mmdata.c
```diff
@@ -1101,6 +1101,39 @@ int nmbrEq(nmbrString *s,nmbrString *t)
...
+/* This simply computes a XOR of the first numbers */
+int nmbrHash(nmbrString *s)
+{
+  static long salt[] = { 4938, 48977, 6897, 7293, 2663, 7925, 2999, 12238,
+      40033, 14038, 10699, 29746, 56108, 34526, 63576, 52053, 61949, 41177, 43740, 22822
+ };
+  long i;
+  long hash = 0;
+  i = -1;
+  while (i < 13 && s[i] != -1) {
```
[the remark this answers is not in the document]
tirix: I'll answer this one now as it is interesting, useful for the rest of the review... and definitely deserves more information in the comments!

Actually, indices -1, -2 and -3 _are_ valid and used in `nmbrString`:
- Index `-1` is the length of the number string. See `nmbrLen` (mmdata.c line 1023)
- Index `-2` is the allocated length, i.e. how many numbers are available totally (could be more than the actual current length of the string). See `nmbrAllocLen` (mmdata.c line 1032)
- Index `-3` is the location in `memUsedPool` (for memory management)

In this specific case, I wanted to include the length of the string in the hash, which I think makes sense but shall have been explained.

In any case, **you clearly spotted a mistake** here because in the salt, index -1 is clearly invalid!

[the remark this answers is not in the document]
digama0: Any use of negative indices should be wrapped behind a macro to provide more safety and clarity. (That is, there should be a macro like `#define nmbrLen(p) p[-1]` where `p` has type `nmbrString` which is a typedef for `int*` or what have you.

wlammen:
New Year's Eve is closing in.  I need some time for it.

mmwsts.c
```diff
@@ -0,0 +1,1069 @@
...
+  case 'c': return "command line";
+  case 'u': return "terminal types";
+  default: return "unknown chunk";
+  }
+}
+
+/* Cache to speed up conversions */
+hashtable stsCache;
+
+/* Math symbol comparison for bsearch */
```
wlammen, line 114: The comment should explain the result, that is limited to -1, 0 and +1, and what is delivered when.
It seems possible that the token list contains the same key multiple times in succession.  Since this is part of a binary search, data may point somewhere in the middle of such a series.  Is it guaranteed to hold just one global element (there is a suspicious active flag in the structure, that may allow a disabled and enabled element in the same series)? Is it guaranteed there is always a global element present?  If either assumption is missed, the binary search may fail.

Clarify in comment.

wlammen (changes requested):

mmwsts.c
```diff
@@ -0,0 +1,1069 @@
...
+  /* Assign mathString */
+  nmbrLet(&mathString, nmbrSpace(mathStringLen));
+  for (i = 0; i < mathStringLen; i++)
+    mathString[i] = wrkNmbrPtr[i];
+  if (mathStringLen) nmbrMakeTempAlloc(mathString); /* Flag for dealloc*/
+  free(wrkNmbrPtr);
+  return mathString;
+}
+
+/* Store a couple key/object into the cache */
```
wlammen, line 195: It is difficult to trace matching alloc/free pairs, if this is silently done in such a function.  Suggest sth like stsMoveToCacheAndClear which indicate that the input variables are cleared.

Document Pre/Postconditions

wlammen (changes requested):

mmwsts.c
```diff
@@ -0,0 +1,1069 @@
...
+}
+
+/* Free a couple key/object from the cache */
+void stsFreeCache(nmbrString **key, vstring *object) {
+  nmbrLet(key, NULL_NMBRSTRING);
+  let(object, NULL);
+}
+
+/* Dump a couple key/object from the cache */
+void stsDumpCache(nmbrString *key, vstring object) {
```
wlammen, line 211: suspicious cast to eqFunc in line 235: The signature of this function does not match that of eqFunc: int (eqFunc)(void *, void *)

mmwsts.c
```diff
@@ -0,0 +1,1069 @@
...
+    return 0;
+  }
+  /* Format is invalid until fully parsed. */
+  let(&stsFormat, "");
+
+  /* Create a cache for exported statements to speed up */
+  if(stsUseCache) {
+    stsCache = htcreate(format, STS_CACHE_BUCKETS, "", (hashFunc*)&nmbrHash, (eqFunc*)&nmbrEq,
+			(letFunc*)&stsStoreCache, (freeFunc*)&stsFreeCache,
+			(eqFunc*)&stsDumpCache);
```
wlammen, line 235: suspicious cast: signature of stsDumpCache does not match eqFunc.

mmwsts.c
```diff
@@ -0,0 +1,1069 @@
...
+}
+
+/* Dump a couple key/object from the cache */
+void stsDumpCache(nmbrString *key, vstring object) {
+//  printf("- %s -> %s\n", nmbrCvtMToVString(key), object);
+  printf("- (%d) %s [%s]\n", nmbrHash(key), nmbrCvtMToVString(key), nmbrCvtAnyToVString(key));
+}
+
+/* Parse a file containing the structured typesetting rules. */
+int parseSTSRules(vstring format) {
```
wlammen, line 217: Coding style:  This function is way too long. It covers more than 300 lines.  This exceeds the recommended max length of 20 lines by far, and its code is easily broken down into steps that can be moved into different helper functions.

document pre/postconditions:  example: stsFormat is both an input and an output variable, but that is not easily seen.

mmwsts.c
```diff
@@ -0,0 +1,1069 @@
...
+  vstring inputFn = "";
+  vstring chunk = "";
+  char chunkType = 0;
+  flag ots = g_outputToString;
+  g_outputToString = 0;
+
+  /* If the same format was already parsed, nothing to do. */
+  if(strcmp(stsFormat, format) == 0) {
```
wlammen, line 225: move the early out code to the beginning of the function where parameter checks usually take place

wlammen (changes requested):

mmdata.c
```diff
@@ -1098,6 +1131,34 @@ nmbrString *nmbrSpace(long n)
   return sout;
 }
 
+/* Search for the nth occurrence of string2 in string1 */
+long nmbrInstrN(long start_position, long occ, nmbrString *string1,
+  nmbrString *string2, long start2, long length2)
+{
+  if (start_position < 1) start_position = 1;
+  start_position--;
+  for(; occ > 0;occ--) {
```
wlammen, line 1140: Optimization: Use a dedicated loop variable in for loops. A compiler will allocate a CPU register for that.
for (long o = occ + 1; --o > 0;) {... (replace occ with o in loop)...}
is how I would write it.

Info: There is a tiny semantic change in my example: If the memory model of the computer supports signed magnitude instead of two's complement AND occ is MAX_LONG then (occ + 1)-1 == occ may not hold. We can safely ignore this nowadays.

mmdata.c
```diff
@@ -1098,6 +1131,34 @@ nmbrString *nmbrSpace(long n)
   return sout;
 }
 
+/* Search for the nth occurrence of string2 in string1 */
+long nmbrInstrN(long start_position, long occ, nmbrString *string1,
```
wlammen, line 1135: why is occ defined as long? Do you really expect more than two billion occurrences (or even 32000 should int be only 16 bit wide) of a substring in a string?

mmdata.c
```diff
@@ -1098,6 +1131,34 @@ nmbrString *nmbrSpace(long n)
...
+    for (i = start_position - 1; i <= ls1 - length2; i++) {
+      flag found = 1;
+      for (j = 0; j < length2; j++) {
+        if (string1[i+j] != string2[start2-1+j]) {
+          found = 0;
+          break;
+        }
+      }
+      if (found) {
+	start_position = i+1;
```
wlammen, line 1153: check indentation

mmhtbl.c
```diff
@@ -0,0 +1,184 @@
...
+#include <stdlib.h>
+#include "mmhtbl.h"
+#include "mminou.h"
+
+/* Static buffer for the linked lists */
+#define NB_LINKEDITEMS 50000L
+#define NO_LINKEDITEM -1
+linked *linkedItems;
+int free_linkedItem;
+flag htinit_done = 0;
```
wlammen, line 18: make this a static variable within htinit().  It is completely private implementation detail in this function.

mmwsts.c
```diff
@@ -0,0 +1,1069 @@
...
+    // In case we don't manage to parse the math string, we shall
+    // revert to token-by-token conversion.
+    let(&mmlLine, asciiMathToTexNoSts(text, statemNum));
+  }
+  nmbrLet(&mathString, NULL_NMBRSTRING);
+  nmbrLet(&orgString, NULL_NMBRSTRING);
+  return mmlLine;
+}
+
+
```
wlammen, line 1025: change this to 
#if 0
test code
#endif
to comply with ANSI C

wlammen (changes requested):

wlammen:
I think there are already lots of ideas and issues for refactoring the source.  In particular extracting code from long functions into auxiliary sub-function can help with further review, since the source becomes a lot more readable then.

In addition merge conflicts have to be resolved.

mmhtbl.c
```diff
@@ -0,0 +1,184 @@
...
+  for(int bucket=0;bucket<hashtable->bucket_count;bucket++) empty_buckets+=hashtable->buckets[bucket] == NO_LINKEDITEM;
+  printf("Hashtable %s: %d entries, load factor=%d%%, %d used buckets, %d%% misses, biggest bucket=%d\n", hashtable->name, hashtable->entries,
+	 htlf(hashtable), (hashtable->bucket_count - empty_buckets), (100-(100*hashtable->founds/hashtable->gets)),
+	 hashtable->biggest_bucket_size);
+}
+
+/* Dumpts the whole table */
+void htdump(hashtable *hashtable) {
+  print2("Hashtable %s:\n", hashtable->name);
+  //for(int bucket=0;bucket<hashtable->bucket_count;bucket++) {
```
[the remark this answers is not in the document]
digama0: I thought we were agreed on C99 now? I don't mind seeing `//` creep in, even if we don't do any bulk conversions.

mmdata.c
digama0: `long` is a signed type, and signed overflow is UB, so I think that this change is legal by the spec.

wlammen: @digama0 Exactly.  Because occ+1 is UB when occ==MAX_LONG, --o may contain anything, and the loop starts with a random value.  This cannot happen with the original code.  We can safely ignore this semantic difference here.

mmhtbl.c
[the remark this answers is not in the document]
wlammen: Have we? Can you point me to where the decision took place?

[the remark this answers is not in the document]
digama0: I'm thinking of https://github.com/metamath/metamath-exe/issues/8#issuecomment-1002937386 . In short: we already compile only on C99, even before taking into account the recent refactors.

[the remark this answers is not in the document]
wlammen: In C99 this issue is not relevant and can be ignored.

tirix:
Thank you very much @wlammen for your efforts reviewing my code!

Indeed now, before this can be merged, I'll have to solve the conflicts with all the refactoring that's going on.

digama0:
I took care of merging this with master. @tirix , you should double check the last commit, which fixes a few warnings I was getting with the original version.

tirix:
Thank you Mario!
This looks good to me!

benjub:
> Thank you Mario! This looks good to me!

@tirix : does that imply this PR can be closed?

tirix:
> @tirix : does that imply this PR can be closed?

This PR was originally made for version 0.181 in December 2021, but much has changed since then and it needed to be rebased. Mario did a first rebase in 2022 and I thanked him for that - but since then there were more changes, resulting in more conflicts to be resolved.

I believe two things are needed for this PR to be merged:
- conflicts must be resolved
- remarks from Wolf have to be addressed.

In the meantime, I've added STS support to [metamath-web](https://github.com/tirix/metamath-web), which is serving [these pages](https://metamath.tirix.org/mpests/toc), so I have not felt urgency to work on this PR. It adds the ability to generate the same pages from `metamath.exe`, which I still believe would be nice to have.

PR #14, opened 2021-12-30
remove support for THINK_C compiler

digama0:
This removes everything that references the `THINK_C` compiler flag.


metamath.c
```diff
@@ -57,7 +56,10 @@
 
 
 
-#define MVERSION "0.198 7-Aug-2021"
+#define MVERSION "0.199.pre 7-Aug-2021"
+/* 0.199.pre
+   30-Dec-2021 mc metamath.c mmdata.c mminou.c mmmaci.c -
+     Remove mmmaci and everything related to THINK_C compiler */
```
digama0, line 62: For future changes, my recommendation is to append them to the `0.199.pre` comment here until the next release. The date is intentionally not updated, to avoid conflicts.

benjub:
There are two other suspiciously small files: mmutil.c/h
Maybe they could be removed as well ?  Here is a discussion I had with Norm a few years ago:

> Similarly,
> it looks like mmutil.c/.h can be deleted, and it looks like "Macintosh
> Think C" is old enough that mmmaci.c/.h may be deleted (after maybe a 
> straw poll on the metamath google group).

Think C was discontinued around 2000 when Apple switched to MacOSX, so
mmmaci.c/.h can be deleted.  An older Mac probably couldn't handle
set.mm anyway.  I guess I haven't touched it partly because of the
saying, "if it ain't broke, don't fix it." :)  One reason it is still
there is that an older C compiler required all files to be named
explicitly, and I didn't want to break some scripts I used to have, but
that is no longer the case.

While mmutil.c is empty now, I had planned to reuse it to break out some
utility functions from other modules that are getting to big, which I
may still do, but I'll think about it.  The idea was to have the empty
module available for future use so I wouldn't have to change the build
scripts I mentioned.

Later:  it looks like David's Makefile lists all modules explicitly.
One more thing to break...

Thanks,
Norm

wlammen:
VAX computers were sold until year 2000 (from the German wiki https://de.wikipedia.org/wiki/Virtual_Address_eXtension).  I suggest to remove the macro VAXC extensions as well.

wlammen (approved):
did a `grep THINK_C *` on the sources.  Count of occurrences matches the deletions here

digama0:
I will remove `mmutil.c` in a separate PR.

jkingdon (approved):

PR #15, opened 2021-12-30
add gitignore, fix c99 warnings

digama0:
There was only one use of `ssize_t` and it wasn't needed, so I've just removed it. It seems to give errors at `-std=c99` for some compilers, including mine (`gcc 9.3.0`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,14 @@
+*.o
+*.Po
+autom4te.cache/
+stamp-h1
+missing
+metamath
+Makefile
+Makefile.in
+compile
+configure
+config.*
+depcomp
+install-sh
+aclocal.m4
--- a/metamath.c
+++ b/metamath.c
@@ -1058,7 +1058,10 @@ void command(int argc, char *argv[]) {
           }
         }
         /* Do the operating system command */
-        (void)system(str1);
+        /* The use of (void)!f() is to ignore the value on both
+          clang (which takes (void) as an ignore indicator)
+          and gcc (which doesn't but is fooled by the ! operator). */
+        (void)!system(str1);
 #ifdef VAXC
         printf("\n"); /* Last line from VAX doesn't have new line */
 #endif
--- a/mminou.c
+++ b/mminou.c
@@ -81,16 +81,6 @@ flag print2(char* fmt, ...) {
   long i;
 
   char *printBuffer; /* Allocated dynamically */
-  /* gcc (Debian 4.9.2-10+deb8u2) 4.9.2 gives error for ssize_t if -c99
-     is specified; gcc (GCC) 7.3.0 doesn't complain if -c99 is specified */
-  /* See https://sourceforge.net/p/predef/wiki/Compilers/ for __LCC__ */
-#ifdef __LCC__   /* ssize_t not defined for lcc compiler */
-  long bufsiz;
-#else
-  ssize_t bufsiz; /* ssize_t (signed size_t) can represent the number -1 for
-                     error checking */
-#endif
-
 
   if (backBufferPos == 0) {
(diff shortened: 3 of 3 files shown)


metamath.c
```diff
@@ -1138,7 +1138,7 @@ void command(int argc, char *argv[])
           }
         }
         /* Do the operating system command */
-        (void)system(str1);
+        (void)!system(str1);
```
wlammen, line 1141: The result of system() is not used.  To suppress warnings, and to indicate this is deliberate, a cast to void is performed. https://stackoverflow.com/questions/689677/why-cast-unused-return-values-to-void

If there is a reason to invert the result this should be noted in a comment.

wlammen (changes requested):

.gitignore
```diff
@@ -0,0 +1,14 @@
+*.o
```
wlammen, line 1: Metamath is not only distributed via GitHub but also via an archive tar file.  In the latter case, a gitignore may overwrite an existing one, because the tar-file contents could be extracted within an existing git environment.

digama0: In this case, we should adjust the archival process to not include the `.gitignore` file (or the `.git` folder, for that matter). Putting the sources in a `src` folder may help with this.

metamath.c
digama0: Will do. The reason is because of https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425 : gcc does not respect the cast-to-void trick, and it is not clear whether this is intentional on their part but it doesn't look like it's going to be fixed any time soon.

An alternative trick is `if(system(str1)) {}`, not sure if that's any better.

.gitignore
digama0: I propose merging this with the .gitignore as is. Currently archive generation is on hiatus since Norm isn't around to keep it running anymore, and when it returns it will be powered by github actions or such and so we will be able to adjust it as necessary.

wlammen: This means we abandon the distribution via a tar file for now.  IMO GitHub is good enough to be the primary (and possibly only) distribution path.  This is still a strategic decision that should be evaluated as an issue first.

metamath.c
wlammen: Fine with me.

.gitignore
digama0: Ok, I'm done for the day but you should open an issue about this. I agree that GitHub can become the new distribution path. The tar file method is already dead I think, since it's manually triggered and 0.198 is the last version that will be produced that way. You can still produce 0.198 from github sources, but 0.200 will use an as-yet-unwritten tar building script running in CI.

metamath.c
jkingdon: If we want a coding style, perhaps we could get ambitious and shoot for "don't ignore the return value from system" (because it could fail if out of memory, out of file descriptors, etc)?

I mean, we have gotten by with this code for a long time, so I'm not saying this is necessarily our top priority, just that.... well putting in a simple error check might not kill us.

jkingdon (approved):

metamath.c
digama0: @jkingdon In this case, the call to `system` is because the user explicitly attempted to call an external program. The program is basically acting like bash here, so failures are expected and should not crash the program, although arguably they should be displayed (even though bash itself doesn't, it just stashes the result in the `$?` variable).

jkingdon: Unless we have something analogous to $? I'd say display a message, yes.

.gitignore
wlammen: I suggest to merge this - there was no real objection #19 after more than a week.

PR #22, opened 2021-12-31
remove obsolete rainbow color models

digama0:
The head comment was slightly adjusted to avoid referencing the now-deleted section.


jkingdon (approved):

benjub:
I think there is a missing "one" in "(i.e., one could...".

Le ven. 31 déc. 2021 à 07:29, Mario Carneiro ***@***.***> a
écrit :

PR #24, opened 2021-12-31
remove mmutil.{c, h}

digama0:
The files are empty.


jkingdon (approved):
We could always bring these back if we wanted them later (although I'd probably advocate for a name/purpose more specific than "util" if we do).

PR #25, opened 2021-12-31
remove INDENT_HTML_PROOFS (make it always on)

digama0:
It's an "experiment" that has been enabled since 2004. I think it was a success.


wlammen (approved):

wlammen:
Count of macro occurrences matches deletions

Issue #17, opened 2021-12-30
Code cleanup


digama0:
This is a todo list for myself or others to tackle, based on my reading of the code. This list is still incomplete, I haven't completed my first pass over the code yet.

* [ ] Rename files to drop `mm` prefix
* [ ] Rename files to use words instead of DOS-compatible 8.3 names
* [ ] Use `#pragma once` instead of include guards. (This is not standardized, but it is [very widely supported](https://en.wikipedia.org/wiki/Pragma_once#Portability).)
* [ ] Raise line limit from 79 characters to 100
* [ ] Delete `VAXC` code (VAX machines are long out of use)
* [ ] Remove `foo(void)` syntax
* [ ] Use `bool`, `true` and `false` instead of `0` and `1`
* [ ] Move variable declarations from top of function to first use
* [ ] Standardize function documentation (doxygen?)
* [ ] Standardize function brace style (I prefer open brace at EOL)
* [ ] Standardize `return(x)` and `return (x)` to `return x`
* [ ] Use `//` comments instead of `/*` (or doc-comments where appropriate)
* [ ] Remove "old buggy compression" [migration code](https://github.com/metamath/metamath-exe/blob/0e6fb2ebf0865fb90354839b2e586490b210d017/mmpars.c#L3381-L3400)
* [ ] `parseProof` returns magic numbers instead of an enum
* [ ] `parseCompressedProof` returns magic numbers instead of an enum
* [ ] Local labels use [encoding](https://github.com/metamath/metamath-exe/blob/0e6fb2ebf0865fb90354839b2e586490b210d017/mmpars.c#L3580-L3581) without macros
* [ ] `edit()` function uses magic numbers
* [ ] `hypReEntryFlagList` uses [magic numbers](https://github.com/metamath/metamath-exe/blob/0e6fb2ebf0865fb90354839b2e586490b210d017/mmpfas.c#L737)
* [ ] `hasFloatingProof` uses `'Y'` and `'N'` instead of true and false
* [ ] `unkSubPrfSteps` uses `'K'` and `'U'` instead of true and false
* [ ] `interactiveUnify` returns magic numbers instead of an enum
* [ ] `readTexDefs` returns magic numbers instead of an enum
* [ ] `processUndoStack` returns defines instead of an enum
* [ ] Remove obsolete flag `g_oldTexFlag`

wlammen:
- Create a best practice instruction file
- Create a maintainers guide.
- replace flag with bool

[...]

PR #30, opened 2022-01-04
Verify markup polarity

### LINKED ISSUE #28: Change two options to "verify markup"
In [PR #2392](https://github.com/metamath/set.mm/pull/2392), the date at the top of set.mm and iset.mm was removed. Therefore, the "verify markup" command that is run by the CI was changed to add the option "skip_top_date" (this was already *meant* to be the case, but it was not: this was fixed in that PR, see the changes in "scripts/verify").

Considering this, it seems natural that the default behavior of "verify markup" should be the one currently obtained with the two options "/top_date_skip /file_skip". These options should should be kept for backward compatibility (although they would have no effect). Additionally, the two options /top_date_chek and /file_check should be added, to restore the current default behavior if needed.

**Edit:** Maybe do this only for /top_date_skip and /top_date_check. The CI still runs the file check (it is only locally that the user generally types /file_skip).

digama0:
* Reverses the polarity of the `gifCheck` argument of `readTexDefs`
* Reverses the polarity of the `fileCheck` argument of `printTexComment`
* Reverses the polarity of the `fileCheck` argument of `writeBibliography`
* Reverses the polarity of the `*Check` arguments of `verifyMarkup`
* The `topDateCheck` and `dateCheck` options are now independent (can be used together in any combination)
* The command line options `/TOP_DATE_SKIP` and `/FILE_SKIP` are now `/TOP_DATE_CHECK` and `/FILE_CHECK` and have the opposite meaning (fixes #28)
* Depends on #18 . (That PR is going to rot quickly, so it would be good to expedite review to avoid too many conflicts with later PRs)


benjub:
That's great ! Looks good to me, although I do not know C (for instance, are ints 0/1 converted to booleans or this kind of things).

Is it possible to convert the list in the comment
> /* 1 = check for missing external files (gifs, bib, etc.) */

to an exhaustive list (e.g, `(gifs, biblio.html, xxx.yyy)`) ?

digama0:
Yes, boolean values are represented as 0/1, and when you use `if (intValue)` it checks if the value is not equal to 0. C99 introduces the `bool`, `true`, `false` constants and one of the items in #17 is using it instead of 0/1, but it hasn't been implemented yet.

metamath.c
```diff
@@ -2130,7 +2130,7 @@ void command(int argc, char *argv[]) {
       }
 
       if (2/*error*/ == readTexDefs(0 /* 1 = check errors only */,
-          0 /* 1 = no GIF file existence check */ )) {
+          1 /* 1 = GIF file existence check */ )) {
```
wlammen, line 2133: The comment should read 0 = ...

wlammen (changes requested):

metamath.c
digama0: No, the meaning of 1 and 0 have flipped here, so `1 = GIF file existence check` now instead of `1 = no GIF file existence check`, and to preserve behavior the actual argument has been changed from 0 to 1 (because in this case we want a GIF check).

mmwtex.h
```diff
@@ -88,7 +88,7 @@ void printTexHeader(flag texHeaderFlag);
 flag printTexComment(vstring commentPtr,    /* Sends result to g_texFilePtr */
     flag htmlCenterFlag, /* 1 = htmlCenterFlag */
     long actionBits, /* see indicators below */
-    flag noFileCheck /* 1 = noFileCheck */);
+    flag fileCheck /* 1 = noFileCheck */);
```
wlammen, line 91: suspicious comment: parameter name does not match semantics in comment

wlammen (changes requested):

mmwtex.h
digama0:
```suggestion
    flag fileCheck /* 1 = fileCheck */);
```

mmcmds.c
```diff
@@ -5233,7 +5233,7 @@ void verifyMarkup(vstring labelMatch,
   /* This code expects a version date at the top of the file such as:
          $( set.mm - Version of 13-Dec-2016 $)
      If we later want a different format, this code should be modified. */
-  if (dateSkip == 0  && topDateSkip == 0) {
+  if (topDateCheck) {
```
wlammen, line 5236: The following code expects mostRecentDate be set in line 5258.  If dateCheck is 0 this seems not be present. Verify.

wlammen (changes requested):

mmwtex.h
wlammen: ack

mmcmds.c
digama0: Actually, although I pushed a bugfix to guard this statement, it looks like it would work anyway (and in particular, this case also comes up when there are no statements in the file), because empty dates are treated as earliest in `compareDates`.

wlammen: I stop looking any further in such corner cases, because the behaviour is obvious, either from the code or a comment, or it is a sloppy implementation, even if it is technically ok.

The presented solution is acked.  Review is finished.

PR #32, opened 2022-01-05
add a documentation draft

wlammen:
Since we need to improve our documentation I provide a suggestion here.  Please explore and comment your findings here.

### DIFF
--- a/doc/Documentation.md
+++ b/doc/Documentation.md
@@ -1,7 +1,9 @@
 # Documentation Rules in Metamath
 
 This is a **draft**.  Here are some links on how to improve on documentation in general:
+
 https://www.oreilly.com/content/the-eight-rules-of-good-documentation/
+
 https://guides.lib.berkeley.edu/how-to-write-good-documentation
 
 1. Documentation is mostly collected in the ``doc`` subfolder of metamath.  As an entry point to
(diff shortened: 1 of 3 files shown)


digama0:
How about using markdown instead of AsciiDoc? They seem to have similar feature sets, but markdown is more well known and in particular gets rendered in github automatically without requiring users to copy the text into https://asciidoclive.com .

wlammen:
@digama0 OK.  A quick lookup on Wikipedia indicates there is possibly no easy transformation into PDF or HTML.  Can that he solved?

digama0:
Markdown to HTML is available in many places, if you want something equivalent to asciidoclive I suggest https://stackedit.io . Searching for markdown to PDF also yields a bunch of results, online services and so on.

wlammen:
In our company we use proprietary software for documentation, so I looked out for something different.  I am not particularly committed to AsciiDoc, anything similar, or even better suits me well.  But should arrive at a decision quickly, so documentation can start.

digama0:
That is, I advocate a documentation style like [this](https://github.com/leanprover-community/lean/tree/master/doc): just loose md files lying about (or collected in a `doc` folder), reachable via links in the readme to make it a kind of low effort static website.

jkingdon:
> That is, I advocate a documentation style like [this](https://github.com/leanprover-community/lean/tree/master/doc): just loose md files lying about (or collected in a `doc` folder), reachable via links in the readme to make it a kind of low effort static website.

If people want to make the case for something else, I'm open to hearing it, but this "directory full of .md files" does strike me as probably the easiest way to "low effort static website".

Issue #31, opened 2022-01-05
Remove tools mode?


digama0:
One fairly large component in metamath.exe is "tools mode", which is a set of text editing utilities borrowed from BASIC reminiscent of unix command line operations like `sed`. Personally, I think that this is best done with another tool rather than bundling it in with the proof assistant and other metamath-specific functionality in metamath.exe. What do people think about removing it? Does metamath tools mode appear in anyone's workflow (or the CI), and if so do we have adequate replacement utilities?

digama0:
This one in particular looks like a maintenance dead weight:
```
MM> HELP UPDATE
This command tags edits made to a program source.  The idea is to keep
all past history of a file in the file itself, in the form of comments.
UPDATE was written for a proprietary language that allowed nested C-style
comments, and it may not be generally useful without some modification.
Essentially a (Unix) diff-like algorithm looks for changes between an
original and a revised file and puts the original lines into the revised
file in the form of comments.  Currently it is not well documented and it
may be easiest just to type UPDATE <return> and answer the questions.
Try it on an original and edited version of a test file to see if you
find it useful.
Syntax:  UPDATE <originfile> <editedinfile> <editedoutfile> <tag> <match>
```

wlammen:
FYI: a quick overview over the range of functions in TOOLS mode.

MM> tools
Entering the Text Tools utilities.  Type HELP for help, EXIT to exit.
TOOLS> help
This utility assists with some common file manipulations.
Most commands will perform an identical operation on each line of a file.
Use HELP ? to see list of help topics.
Note:  When an output file is created, any previous version is renamed,
with ~1 appended, and any ~1 renamed to ~2, etc. (up to ~9, which is lost).
Note:  All string-matching command arguments are case-sensitive.

Line-by-line editing commands:
  ADD - Add a specified string to each line in a file
  CLEAN - Trim spaces and tabs on each line in a file; convert characters
  DELETE - Delete a section of each line in a file
  INSERT - Insert a string at a specified column in each line of a file
  SUBSTITUTE - Make a simple substitution on each line of the file
  TAG - Like ADD, but restricted to a range of lines
  SWAP - Swap the two halves of each line in a file
Other file processing commands:
  BREAK - Break up (parse) a file into a list of tokens (one per line)
  BUILD - Build a file with multiple tokens per line from a list
  COUNT - Count the occurrences in a file of a specified string
  NUMBER - Create a list of numbers
  PARALLEL - Put two files in parallel
[...]

PR #34, opened 2022-01-06
Add tools-mode shell scripts

digama0:
This is a start on the plan in https://github.com/metamath/metamath-exe/issues/31#issuecomment-1006181485, providing bash scripts with functionality equivalent to the correspondingly named TOOLS commands.

The list of utilities is mostly complete, with the following exceptions:
* `SUBSTITUTE` - supported for `oldStr` not crossing a line boundary
* `CLEAN` - fully supported except for the `T (tab)` and `U (untab)` options. Considering that these are hardcoded to 8-column tab stops I'm guessing no one uses these anyway.
* `UPDATE` - not supported
* `COUNT` - not supported, because it doesn't seem useful for scripting as written.
* `SORT` - supported only for empty key (where it is equivalent to the unix `sort` command). The key version seems complicated and not worth supporting unless there is a use case.
* `TOOLS`, `EXIT`/`QUIT` - these are sometimes used in tools command scripts to go in and out of tools mode. This is not supported, since we can't "exit to metamath mode" since we aren't actually in metamath.

Most of the scripts are written in awk, which handles these line-oriented edit operations well. The scripts follow the usual unix design of working as pipes from input to output, so the equivalent of `ADD foo.txt 'a' 'b'` is `tools/add.sh 'a' 'b' < foo.txt > foo2.txt`. However, this does not work when the input is the same as the output (as in most of the metamath commands), and the `inplace.sh` command can be used to buffer the file, so that `inplace.sh foo.txt add.sh 'a' 'b'` acts like one would expect.

I think these tools are useful regardless of the outcome of #31, but one downside of these scripts compared to metamath.exe is that they are not as cross-platform - they work on anything POSIX-ish but not windows (at least, not without a compatibility layer).


tools/add.sh
```diff
@@ -0,0 +1,18 @@
...
+This command adds a character string prefix and/or suffix to each
+line in a file.
+Syntax:  add.sh BEGSTR ENDSTR < INFILE > OUTFILE
+HELP
+}
+
+if [ $# -eq 1 ] && [ "$1" = "-h" ]; then usage; exit; fi
+if [ $# -ne 2 ]; then usage; exit 1; fi
+
+while read line; do echo $1${line}$2; done
```
wlammen, line 18: The script won't stop on
./add.sh a b

wlammen (changes requested):

tools/add.sh
digama0: You must provide two parameters, plus you should pipe it the input file if you don't want it to sit there doing nothing. To skip one or the other, you pass an empty string, i.e. `add.sh pfx "" < in.txt` will prepend `pfx` on every line. I believe this matches the metamath behavior, except that it will go through an interactive prompt to get you to finish the invocation if you don't pass enough parameters the first time.

digama0: As mentioned in the PR comment, these are all "pipe" style programs like `cat`, which is useful for chaining but means that using the same file as input and output does not work. There is an additional utility `inplace.sh` provided for this purpose: `inplace.sh in.txt add.sh pfx ""` will call `add.sh pfx ""` with an in-memory copy of `in.txt` and write the result back to `in.txt`.

tools/delete.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
```
wlammen, line 1: Verify this is the intended result:
echo '12aaa34aaa56' > xx
./delete.sh 'aa' 'aa' < xx > yy
cat yy
12a56

tools/add.sh
wlammen: suggest to remove the or between prefix and suffix and be more explicit about BLANK

tools/delete.sh
digama0: At least, that matches the behavior I intended for it: the second match starts where the first match ended, so they can't overlap.

wlammen: delete str str deletes the first occurrence of str now, a corner case with simple semantics for which I can envision applications.

tools/clean.sh
```diff
@@ -0,0 +1,109 @@
...
+for var in "$@"; do
+  case $var in
+    D | d) allDiscard=1;;
+    B | b) leadDiscard=1;;
+    E | e) trailDiscard=1;;
+    R | r) reduce=1;;
+    Q | q) quote=1;;
+    T | t) echo "T (tab) flag unsupported"; exit 2;;
+    U | u) echo "U (untab) flag unsupported"; exit 2;;
+    P | p) trim=1;;
```
wlammen, line 41: I had to get back here to lookup what trim means. Suggest to rename it to clearParity

tools/clean.sh
```diff
@@ -0,0 +1,109 @@
...
+  for (i=1;i<=NF;i++) {
+    if (quote && ($i == "\"" || $i == "\x39"))
+      inQuote = !inQuote;
+    if (inQuote) {
+      out=out $i;
+      continue;
+    }
+    if (allDiscard && ($i == " " || $i == "\t")) continue;
+    if (trim) $i = chr[and(ord[$i], 0x7f)];
+    if (discardCtrl && discardArr[$i]) continue;
```
wlammen, line 98: optimization: move this one line up

tools/clean.sh
```diff
@@ -0,0 +1,109 @@
...
+    if (caseFlag && ord[$i] < 128) $i = toupper($i);
+    if (lowercase && ord[$i] < 128) $i = tolower($i);
+    if (screen && gfxArr[$i]) $i = gfxArr[$i];
+    out=out $i;
+  }
+  if (trailDiscard)
```
wlammen, line 104: I wonder whether one can have this at the beginning: it eliminates checks in the for loop

wlammen (changes requested):

tools/add.sh
wlammen: ack

tools/delete.sh
wlammen: ack

tools/inplace.sh
```diff
@@ -0,0 +1,22 @@
...
+designed for piping and generally do not work correctly when the same
+file is chosen as input and output.
+HELP
+}
+
+if [ $# -lt 2 ]; then usage; exit 1; fi
+
+file=$1; shift
+buf=$(cat $file)
+echo "$buf" | $@ > $file
```
wlammen, line 22: review ok

tools/insert.sh
```diff
@@ -0,0 +1,28 @@
...
+in a file.  It is intended to aid further processing of column-
+sensitive files.  Note: the index of the first column is 1, not 0.  If a
+line is shorter than COLUMN, then it is padded with spaces so that
+STRING is still added at COLUMN.
+Syntax:  insert.sh STRING COLUMN < INFILE > OUTFILE
+HELP
+}
+
+if [ $# -eq 1 ] && [ "$1" = "-h" ]; then usage; exit; fi
+isNum() { case "$1" in '' | '0' | *[!0-9]*) return 1;; esac; }
```
wlammen, line 19: 00 isAllowed? verify

wlammen (changes requested):

tools/insert.sh
digama0: it is, the error checking isn't air tight here but I don't think it's necessary to be. If you pass 0 then it treats it as 1. (Probably passing huge numbers will fail too.) I'm not sure what kind of environment (i.e. what bash-isms) I'm allowed to assume of the user here. If you have a better suggestion feel free to improve it.

tools/clean.sh
digama0: That does not preserve behavior: if `trim` and `discardCtrl` are both enabled, then `0x88` would be discarded with the current order but not if the discard precedes the trim. This is a direct transcription from the `edit()` function in metamath so I would want to change it there too if so.

tools/clean.sh
digama0: That kind of motion is not likely to preserve behavior, since the for loop can do quite a few things to the input

digama0:
This PR is now done (to my satisfaction). The `translate.sh` command can be used to turn a tools script into a bash script, and `submit.sh` uses it to run tools commands. Like the other commands in the suite, `submit.sh` takes its input via stdin, which means that you can also use it interactively like `bash`, since it will translate each command and execute it as soon as it receives it. You can use `translate.sh` to translate a tools file into a bash file (it will preserve comments), or you can use `submit.sh` to translate a tools file on the spot and execute it.

One drawback of using shell scripts is that they have to be located somewhere, and `submit.sh` needs to know where the other shell scripts are (including itself). For that purpose, it takes an argument which is the `tools/` directory. If you leave it off, it will default to `mm-tools`, with the idea that you can symlink `mm-tools` on your path and then calls to `mm-tools/add.sh` will work as expected, which would make more sense for scripts like `min.cmd` which are checked in and not user-specific.

tools/clean.sh
wlammen: There is a deeper problem here: If the parity bit is used to ensure proper transmission from an unreliable source, this parity bit clearance has to be executed first.  The space character SP 0x20, for example, is transmitted as 0xA0 if parity must be even.  This immediately busts the checks in line 86 and 96.  Bad.  So the parity bit has highest priority and line 97 must be moved upwards to line 90.  And line 86 is in a wrong position, too.  Why should leading spaces be handled differently from trailing ones?  If any of the transformations in the loop is capable of producing a space,  then the first character may be converted into a leading space,  which might be prohibited, and the algorithm fails again.

To get back to your remark: option P and V are mutually exclusive.  And who on earth cares about teletyping and non-graphical monitors any more?  So lets leave this issue at rest.  These options are eligible to removal.

Keeping the code for leading and trailing spaces in different locations still looks awkward.  But spaces or tabs are not produced by any transformation, so the implementation is somewhat risky, but not exactly failing.

That said I leave it up to you whether this is corrected.

tools/clean.sh
wlammen: OK, but then move the check for leading spaces here, too.

tools/beep.sh
```diff
@@ -0,0 +1,19 @@
...
+running command has started, it will alert you that the command is
+finished.
+Syntax: beep.sh
+HELP
+}
+
+if [ $# -eq 1 ] && [ "$1" = "-h" ]; then usage; exit; fi
+if [ $# -ne 0 ]; then usage; exit 1; fi
+
+beep
```
wlammen, line 19: Review ok

tools/clean.sh
digama0: Like you say, most of the features here are significantly out of date and useless. My main interest is in preserving the behavior of the `edit()` function as written, at least for those options that can be simply implemented. I would very much like to remove the function entirely (from metamath) at some point, but that relies on having an appropriate fallback for applications that are still using it for some reason. So even though the behavior of this function is totally nonsensical when it comes to parity bit mangling (when was the last time the 8th bit was parity anyway?), I'm inclined to keep it as is for compatibility's sake.

wlammen: The parity flag is still present on super-modern Intel CPUs.  Its an heritage of the venerable 8086.  This roughly gives an idea when parity was used in transmission protocols

mmhlpa.c
```diff
@@ -157,7 +158,8 @@ H("Note: The SUBSTITUTE command may be abbreviated by S.");
 
 g_printHelp = !strcmp(saveHelpCmd, "HELP SWAP");
 H("This command swaps the parts of each line before and after a");
-H("specified string.");
+H("specified string <middle>.");
+H("Syntax:  SWAP <iofile> <middle>");
 
 
 g_printHelp = !strcmp(saveHelpCmd, "HELP INSERT");
```
wlammen, line 165: checked all occurrencies of 'parity'.  Review ok.

mmvstr.c
```diff
@@ -384,13 +384,13 @@ vstring edit(vstring sin,long control) {
...
        sout[i] = toupper(sout[i]);
     */
     /* 13-Jun-2009 nm The upper/lower case C functions have odd behavior
        with characters > 127, at least in lcc.  So this was rewritten to
        not use them. */
-    if ((case_flag) && (sout[i] >= 'a' && sout[i] <= 'z'))
+    if ((uppercase_flag) && (sout[i] >= 'a' && sout[i] <= 'z'))
        sout[i] = (char)(sout[i] - ('a' - 'A'));
 
     /* Convert [] to () */
```
wlammen, line 396: All former case_flag locations converted.  Review ok

tools/break.sh
```diff
@@ -0,0 +1,41 @@
...
+
+if [ "$1" = "" ]; then specChars="()[],=:;{}"; else specChars=$1; fi
+
+awk -v specChars="$specChars" \
+'BEGIN {
+  len = length(specChars);
+  split(specChars, specCharsArr, "");
+} {
+  for (i = 1; i <= len; i++) {
+    if (!index($0, specCharsArr[i])) continue;
```
wlammen, line 27: I think the implementation of the loop works. But is
```
escaped = specCharsArr[i];
sub(/[\\\"]/, "\\" specCharsArr[i], escaped);
sub(escaped, "  " specCharsArr[i] " ");
```
not more effective?

wlammen (changes requested):

tools/break.sh
```diff
@@ -0,0 +1,41 @@
...
+# Mario Carneiro 7-Jan-2022
+
+usage() {
+  cat << HELP
+BREAK - Break up (parse) a file into a list of tokens (one per line)
+
+This command breaks up a file into tokens, one per line, breaking at
+whitespace and any special characters you specify as delimiters.
+Use an explicit (quoted) space as SPECCHARS to avoid the default
+special characters and break only on whitespace.
```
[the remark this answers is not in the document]
digama0: LF can be included in the list of delimiters, but it doesn't do anything as there is already a newline there. Perhaps `"\n"` would be a better suggestion than `" "` for turning off the default specChars.

tools/break.sh
digama0: You would still have to traverse over the matched character to avoid an infinite loop. I believe that something like this can work, but it's not obvious to me that it is shorter. It would be nice if there were a way to avoid the loop too and do it all with one regex.

wlammen: the 3rd line replaces most of your while loop. I don't want to be nitpicky here. It is several years ago that I wrote my last awk script, and I forgot 80% or more about the syntax and efficient coding patterns. Eventually I noticed that I never saw the match shorthand & in replacement strings, and here is an example of how to use it.

tools/build.sh
```diff
@@ -0,0 +1,31 @@
...
+# Shell script equivalent of Metamath BUILD tool
+# Mario Carneiro 7-Jan-2022
+
+usage() {
+  cat << HELP
+BUILD - Build a file with multiple tokens per line from a list
+
+This command combines a list of tokens into multiple tokens per line,
+as many as will fit per line, separating them with spaces.
+Syntax:  build.sh < INFILE > OUTFILE
```
wlammen, line 11: Suggest to add: A line is terminated after it became more than 72 characters wide.

wlammen (changes requested):

tools/copy.sh
```diff
@@ -0,0 +1,24 @@
...
+previously exist in 1.tmp.
+Syntax:  copy.sh OUTFILE INFILE INFILE ...
+HELP
+}
+
+if [ $# -eq 1 ] && [ "$1" = "-h" ]; then usage; exit; fi
+
+outfile=$1; shift
+buf=$(cat $@)
+echo "$buf" > "$outfile"
```
wlammen, line 24: Review ok

tools/duplicate.sh
```diff
@@ -0,0 +1,19 @@
...
+This command finds all duplicate lines in a file and places them, in
+sorted order, into the output file.
+Syntax:  duplicate.sh < INFILE > OUTFILE
+HELP
+}
+
+if [ $# -eq 1 ] && [ "$1" = "-h" ]; then usage; exit; fi
+if [ $# -ne 0 ]; then usage; exit 1; fi
+
+sort | uniq -d
```
wlammen, line 19: suggest using option -u with sort instead. https://stackoverflow.com/questions/3382936/sort-uniq-in-linux-shell

wlammen (changes requested):

tools/duplicate.sh
wlammen, line 19: Review ok

tools/inplace.sh
```diff
@@ -0,0 +1,22 @@
...
+designed for piping and generally do not work correctly when the same
+file is chosen as input and output.
+HELP
+}
+
+if [ $# -lt 2 ]; then usage; exit 1; fi
+
+file=$1; shift
+buf=$(cat $file)
+echo "$buf" | "$@" > $file
```
wlammen, line 22: Review ok

tools/build.sh
digama0: It's actually 73 characters, because of the space insertion. I think this has to do with the recommended line length in git commit messages, but I can't say for sure.

wlammen: ack

tools/break.sh
wlammen: This eliminates the whole for loop and makes the split command in BEGIN redundant
```
    escaped = specChars;
    # metachars need to be escaped to work as a dynamic regexp
    gsub(/[\\\]\-\^]/, "\\\\&", escaped);
    escaped = "[" escaped "]";
    print escaped;
    gsub(escaped, " & ");
```

digama0: updated

wlammen: ack

tools/insert.sh
```diff
@@ -0,0 +1,28 @@
...
+isNum() { case "$1" in '' | *[!0-9]*) return 1;; esac }
+if [ $# -ne 2 ] || ! isNum "$2" || [ $2 -eq 0 ]; then usage; exit 1; fi
+
+awk -v insertStr="$1" -v len=$2 '
+BEGIN {
+  gsub("%", "%%", insertStr);
+  fmtStr = "%-" (len - 1) "s" insertStr "%s\n";
+} {
+  printf(fmtStr, substr($0, 1, len - 1), substr($0, len));
+}'
```
wlammen, line 28: Review ok

tools/match.sh
```diff
@@ -0,0 +1,22 @@
...
+}
+
+if [ $# -eq 1 ] && [ "$1" = "-h" -o "$1" = "--help" ]; then usage; exit; fi
+if [ $# -ne 2 ]; then usage; exit 1; fi
+
+case "$2" in
+  Y | y)  grep -F "$1";;
+  N | n)  grep -v -F "$1";;
+  *)  usage; exit 1;;
+esac
```
wlammen, line 22: review ok

tools/number.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
wlammen, line 1: suggest to let the bash command seq do the work

wlammen (changes requested):

tools/parallel.sh
```diff
@@ -0,0 +1,19 @@
...
+The two files should have the same number of lines; if not, the longer
+file is paralleled with empty strings at the end.
+Syntax:  parallel.sh INFILE1 INFILE2 BTWNSTR > OUTFILE
+HELP
+}
+
+if [ $# -eq 1 ] && [ "$1" = "-h" -o "$1" = "--help" ]; then usage; exit; fi
+if [ $# -ne 3 ]; then usage; exit 1; fi
+
+paste -d "$3" "$1" "$2"
```
wlammen, line 19: Review ok

tools/reverse.sh
```diff
@@ -0,0 +1,17 @@
...
+
+This command reverses the order of the lines in a file.
+Syntax:  reverse.sh < INFILE > OUTFILE
+HELP
+}
+
+if [ $# -eq 1 ] && [ "$1" = "-h" -o "$1" = "--help" ]; then usage; exit; fi
+if [ $# -ne 0 ]; then usage; exit 1; fi
+
+tac
```
wlammen, line 17: Review ok

tools/number.sh
digama0: Strictly speaking `seq` doesn't have the same behavior as `NUMBER` on negative values of incr (`seq` does the useful thing and `NUMBER` stops after 1 step). But I guess that means no one is using the behavior anyway...

tools/right.sh
```diff
@@ -0,0 +1,25 @@
...
+if [ $# -eq 1 ] && [ "$1" = "-h" -o "$1" = "--help" ]; then usage; exit; fi
+if [ $# -ne 0 ]; then usage; exit 1; fi
+
+buf=$(cat)
+len=$(echo "$buf" | awk '
+  BEGIN { max = 0; }
+  { if (length($0) > max) max = length($0); }
+  END { print max }
+')
+echo "$buf" | awk -v len=$len '{ printf "%" len "s\n", $0 }'
```
wlammen, line 25: review ok

tools/number.sh
digama0: Not sure if you are putting this comment on the shebang line to point it out, but as an FYI, this one uses `bash` instead of `sh` because it needs the `[[ "$1" == ?(-)+([0-9]) ]]` command which doesn't work in `sh`. Maybe you have a better idea how to test for numerals in posix compliant mode?

tools/sort.sh
```diff
@@ -0,0 +1,21 @@
...
+If the key string is blank, the line is compared starting at column 1.
+If a line doesn't contain the key, it is compared starting at column 1.
+Syntax:  sort.sh KEY < INFILE > OUTFILE
+HELP
+}
+
+if [ $# -eq 1 ] && [ "$1" = "-h" -o "$1" = "--help" ]; then usage; exit; fi
+if [ $# -ne 1 ]; then usage; exit 1; fi
+
+if [ "$1" != "" ]; then echo "SORT with key unsupported"; exit 2; fi
```
wlammen, line 19: Why is key not supported? Difficult to write? No usage known?

tools/number.sh
wlammen: check out here:
https://stackoverflow.com/questions/4137262/is-there-an-easy-way-to-determine-if-user-input-is-an-integer-in-bash

tools/sort.sh
digama0: Difficult to write. There isn't an easy way to do it with `sort` AFAICT, so it would have to be a python script or something like it.

tools/swap.sh
```diff
@@ -0,0 +1,22 @@
...
+}
+
+if [ $# -ne 1 ]; then usage; exit 1; fi
+
+awk -v middle="$1" '{
+  start = index($0, middle);
+  if (start)
+    $0 = substr($0, start + length(middle)) middle substr($0, 1, start - 1);
+  print;
+}'
```
wlammen, line 22: review ok

tools/unique.sh
```diff
@@ -0,0 +1,18 @@
...
+This command finds all unique lines in a file and places them, in
+sorted order, into the output file.
+Syntax:  unique.sh < INFILE > OUTFILE
+HELP
+}
+
+if [ $# -eq 1 ] && [ "$1" = "-h" -o "$1" = "--help" ]; then usage; exit; fi
+if [ $# -ne 0 ]; then usage; exit 1; fi
+
+sort | uniq -u
```
wlammen, line 18: review ok

tools/unduplicate.sh
```diff
@@ -0,0 +1,17 @@
...
+
+This command sorts a file then removes any duplicate lines from the output.
+Syntax:  unduplicate.sh < INFILE > OUTFILE
+HELP
+}
+
+if [ $# -eq 1 ] && [ "$1" = "-h" -o "$1" = "--help" ]; then usage; exit; fi
+if [ $# -ne 0 ]; then usage; exit 1; fi
+
+sort -u
```
wlammen, line 17: review ok

tools/type.sh
```diff
@@ -0,0 +1,28 @@
...
+  head -10
+elif [ $# -eq 1 ]; then
+  case "$1" in
+    a* | A*)  cat; exit;;
+    '' | *[!0-9]*)  usage; exit 1;;
+    *)  head -$1;;
+  esac
+else
+  usage; exit 1
+fi
```
wlammen, line 28: review ok

tools/number.sh
wlammen: Why are we concerned with a Unix shell, but not bash?  translate.sh explicitly refers to bash.  Lets downsize in a different  PR and assume bash for now.

tools/sort.sh
digama0: I implemented something that mostly works, as long as you aren't sorting something containing the nonprintable character `\x01`.

tools/number.sh
digama0: Works for me. Writing portable shell scripts is a headache.

tools/sort.sh
```diff
@@ -0,0 +1,27 @@
...
+
+if [ "$1" == "" ]; then
+  sort
+else
+  awk -v matchKey="$1" '{
+    n = index($0, matchKey);
+    key = n ? substr($0, n) : $0;
+    print key "\1" $0;
+  }' | sort -k1,1 -t $'\x01' | sed 's/.*\x01//'
+fi
```
wlammen, line 27: Nice.
A slightly better choice of the delimiter is 0x1F, see here https://stackoverflow.com/questions/492090/least-used-delimiter-character-in-normal-text-ascii-128.  This could be the default for an optional second parameter of sort.sh

tools/tag.sh
```diff
@@ -0,0 +1,47 @@
...
+HELP
+}
+
+if [ $# -eq 1 ] && [ "$1" = "-h" -o "$1" = "--help" ]; then usage; exit; fi
+isNum() { case "$1" in '' | *[!0-9]*) return 1;; esac; }
+if [ $# -ne 6 ] || ! isNum "$4" || ! isNum "$6"; then usage; exit 1; fi
+
+awk -v before="$1" -v after="$2" -v start="$3" -v end="$5" \
+  -v startNum=$4 -v endNum=$6 '{
+  if (startNum > 0) {
```
wlammen, line 38: startnum == 0 is not a valid input according to the description.  Here it is used as if it starts the range.

wlammen (changes requested):

tools/tag.sh
digama0: `startNum` changes during execution, it counts down to zero. I don't think an initial setting of `startNum == 0` is exceptional in any way code-wise, but it would be possible (with some additional code) to specifically exclude it if required.

wlammen: I find it confusing to deal with undocumented input.  Instead of proper error handling I will see some obscure and unexpected  behaviour.  Once I detect that start == 0 means 'from the beginning' I am lured into thinking end == 0 means 'to the end'.  And so on.  Let the script do what the help says it does.  The rest is just confusing.

tools/translate.sh
```diff
@@ -0,0 +1,138 @@
...
+
+if [ $# -eq 0 ]; then tools="mm-tools"
+elif [ $# -eq 1 ]; then tools=$1
+else usage; exit 1; fi
+
+echo "#!/bin/bash"
+awk -v tools="$tools" -v echo="$echo" -v recursive="$recursive" \
+'BEGIN {
+  len = length(specChars);
+  split(specChars, specCharsArr, "");
```
wlammen, line 40: where is this used?

tools/tag.sh
digama0: What I mean to say is that I did not think about the start == 0 case at all when writing this (and this is also just a copy of the metamath code logic), so it does whatever the logical extension of the > 0 behavior is.

tools/translate.sh
```diff
@@ -0,0 +1,138 @@
...
+  if (recursive) {
+    extraArg = tools;
+    gsub(/[\\$"]/, "\\\\&", extraArg);
```
wlammen, line 43: according to the bash manual, the apostrophe ' needs quoting as well.

wlammen (changes requested):

tools/translate.sh
digama0: You sure? `echo "I'm testing quotes"` seems to work just fine

tools/translate.sh
digama0: It's not.

tools/translate.sh
wlammen: From the bash manual:
3.1.2.3 Double Quotes
Enclosing characters in double quotes (‘"’) preserves the literal value of all characters within
the quotes, with the exception of ‘$’, ‘‘’, ‘\’, and, when history expansion is enabled, ‘!’.

digama0: Although I see that text in the manual, I also can't find any special meaning for `'` outside of its interaction with `${...}` which doesn't matter if all `$` are quoted. The history expansion thing looks worrisome but based on my testing it's not active when reading from scripts.

digama0: Oh! That character isn't the apostrophe, it's a backtick. ``"`command`"`` can be used to execute a command and insert the result into the string, and indeed it needs to be escaped.

tools/translate.sh
```diff
@@ -0,0 +1,136 @@
...
+
+if [ "$1" = "-r" ]; then recursive=1; shift; fi
+if [ "$1" = "-e" ]; then echo=1; shift; fi
+
+if [ $# -eq 0 ]; then tools="mm-tools"
+elif [ $# -eq 1 ]; then tools=$1
+else usage; exit 1; fi
+
+echo "#!/bin/bash"
+awk -v tools="$tools" -v echo="$echo" -v recursive="$recursive" \
```
wlammen, line 37: compile error
cat xx
exit
wolf@wolf-lammen:~/metamath/src/review/sts/tools$ ./translate.sh "/home/wolf/metamath/src/review/sts/tools" < xx > yy
awk: line 15: syntax error at or near {
awk: line 17: syntax error at or near :
awk: line 18: syntax error at or near :
awk: line 19: syntax error at or near :
awk: line 20: syntax error at or near :

wlammen (changes requested):

tools/translate.sh
wlammen: line 64 is also affected

tools/translate.sh
digama0: Hm, I can't replicate. Based on the position of the error, it sounds like your awk version doesn't support `switch`? If you can confirm, I'll rewrite it to avoid that.

wlammen: awk -W version
mawk 1.3.4 20200120
Copyright 2008-2019,2020, Thomas E. Dickey
Copyright 1991-1996,2014, Michael D. Brennan

random-funcs:       srandom/random
regex-funcs:        internal
compiled limits:
sprintf buffer      8192
maximum-integer     2147483647

wlammen: It seems it does not support a switch statement. This won't compile:
awk -v tools="$tools" -v echo="$echo" -v recursive="$recursive" \
'{
    x = $0
    switch (x)
    {
      case "x":
        print
    }
}'

tools/translate.sh
```diff
@@ -0,0 +1,120 @@
...
+      $1 = tools "/inplace.sh " infile;
+    } else if ($1 == "HELP" || $1 == "COUNT" || $1 == "UPDATE") {
+      $0 = "# " $1 " command not supported";
+    } else {
+      gsub(/"/, "\\\\&");
+      $0 = "# unknown command: \"" $0 "\"";
+    }
+    if (echo) {
+      echoLine = $0;
+      gsub(/[\\$"]/, "\\\\&", echoLine);
```
wlammen, line 114: needs backtick escape as well?

wlammen (changes requested):

tools/tag.sh
wlammen: ack

tools/substitute.sh
```diff
@@ -0,0 +1,41 @@
...
+  A | a)  all=1;;
+  '' | '0' | *[!0-9]*)  usage; exit 1;;
+  *)  occs=$3;;
+esac
+
+# multiline oldStr not supported
+awk -v oldStr="$1" -v newStr="$2" -v matchStr="$4" \
+  -v all="$all" -v occs=$occs \
+'BEGIN {
+  gsub(/[][\\.^$(){}|*+?]/, "\\\\&", oldStr);
```
wlammen, line 33: the caret ^ needs to be escaped

tools/substitute.sh
```diff
@@ -0,0 +1,41 @@
...
+  gsub(/[&\\]/, "\\\\&", newStr);
+} {
+  if (index($0, matchStr)) {
+    if (all) gsub(oldStr, newStr);
+    else gensub(oldStr, newStr, occs + 1);
```
wlammen, line 38: gensub is not defined in mawk

wlammen (changes requested):

tools/translate.sh
```diff
@@ -0,0 +1,120 @@
...
+      print "echo \"" echoLine "\"";
+    }
+    file = tools "/" tolower($1) ".sh";
+    if (simpleCommands[$1]) {
+      infile = $2;
+      $2 = file;
+      $1 = tools "/inplace.sh " infile;
```
wlammen, line 82: The idea of a simple command is that it follows the syntax rule
```<cmd> <file> <other args>```
in metamath scripts.
This is an implicit assumption here.  Make it explicit by adding a comment to this respect in line 79 or 44

**Review by wlammen — CHANGES_REQUESTED** [review]

File: tools/sort.sh [hunk 37]
```diff
@@ -0,0 +1,27 @@
...
+
+usage() {
+  cat << HELP
+SORT - Sort the lines in a file with key starting at specified string
+
+This command sorts a file, comparing lines starting at a key string.
+If the key string is blank, the line is compared starting at column 1.
+If a line doesn't contain the key, it is compared starting at column 1.
+Syntax:  sort.sh KEY < INFILE > OUTFILE
+HELP
```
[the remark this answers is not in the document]
digama0: I'd rather not ackowledge it in the docs, it's not part of the spec. The right solution is to implement this in python or other regular programming language as I mentioned, this is just a limitation of the quick and dirty solution that hopefully will not interfere with standard text manipulation activities.

tools/translate.sh
digama0: It's not an assumption, it's a classification of the scripts. It just so happens that most of the TOOLS commands have this form, so they can be handled in one go. It's not a user facing classification at all, and it is explicitly enumerated so there are no implicit assumptions here. Maybe the name could be better, but it doesn't seem very important.

tools/substitute.sh
digama0: Only if it starts the bracket list.

tools/sort.sh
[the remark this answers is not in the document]
wlammen: ..hopefully...  That is precisely what my comment is about.  Mark it as an implementation detail, workaround or todo, but mention it.  It describes the behaviour.  For the overall concept see here https://stackoverflow.com/questions/35303332/what-are-preconditions-and-postconditions

tools/substitute.sh
```diff
@@ -0,0 +1,55 @@
...
+} {
+  if (index($0, matchStr)) {
+    if (all) gsub(oldStr, newStr);
+    else {
+      # non-gawk version of:
+      # $0 = gensub(oldStr, newStr, occs + 1);
+
+      lhs = "";
+      for (i = occs; i > 1; i--) {
+        if (!match($0, oldStr)) break;
```
wlammen, line 45: looks like this line belongs to the condition of the for statement

wlammen (changes requested):

tools/sort.sh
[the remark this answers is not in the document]
digama0: Well, this is why the original version was throwing an error, because I didn't want to deal with this. It's tricky to throw an error in the middle of an awk script, but I put something that should work well enough.

tools/translate.sh
wlammen: I tested this block of code, naively entering the parameters in the order required by the script usage help
echo 'add x y dummy-file-name' | ./translate.sh path-to-tools
and of course it failed.

What is the reason behind this failure? I never wrote a metamath script, so I was completely unaware of the grammar details of this language. My assumption was, it parallels that of the bash scripts.  Which is not exactly true.

In the same manner this program (or chunk of code) assumes a particular format of the input script.  This is part of the contract (Design by contract https://en.wikipedia.org/wiki/Design_by_contract) between the user of this script and you as a programmer.  As with real life contracts, their clauses should be noted somewhere.  And somewhere does not mean "elsewhere", but either here, or at least referenced here.

This contract thinking is the result of ages of sloppy programming.  It counteracts situations like that in the beginning of this post.  And tedious debugging sessions.

My two cents

digama0: I'm confused. The documentation of `translate.sh` says that it is meant to translate TOOLS scripts, so it uses the syntax for tools scripts (and indeed it can't do anything else). The whole point of it is to translate between the metamath argument order and syntax conventions to bash syntax and conventions. If you want to use the shell files directly then just do that can't don't call `translate.sh`.

If you want to know what the grammar details of the metamath TOOLS language are, you should consult the metamath TOOLS help. I suppose I could reproduce the entirety of that documentation inside the help of `translate.sh`, but the main purpose of this command is for "modernizing your old scripts" so the presumption is that you already know the TOOLS syntax or else you have a file written previously with knowledge of this syntax and are interested in converting it.

You are preaching to the choir re: design by contract, but metamath.exe is currently in a shambles with respect to this criterion and it will take a while to get to that point. The purpose of these scripts is to get TOOLS out of metamath.exe and replace them with something good enough. They are *not* air tight regarding error handling for a bunch of reasons, they *will* break if you push on them too hard. The documentation on the scripts is passable but not comprehensive. (If you want to make a documentation suggestion, just post it.) And they mostly either do things that already exist as unix commands or do a weird variation that is not generally useful. In short, I wouldn't recommend anyone use them for new scripts, they exist only for compatibility purposes.

PS: I mentioned earlier that metamath uses low ascii without acknowledging its limitations. Just knowing that this is how the program works makes it easy to trigger bugs:

```bash
$ echo -e '\x03' > test.mm
$  ./metamath test.mm
Metamath - Version 0.199.pre 7-Aug-2021       Type HELP for help, EXIT to exit.
MM> READ "test.mm"
Reading source file "test.mm"... 2 bytes

?Error on line 1 of file "test.mm":
?BUG CHECK:  *** DETECTED BUG 1514
...
```

tools/substitute.sh
wlammen: From gawk manual section 3.4 bracket expression p. 54
To include one of the characters ‘\’, ‘]’, ‘-’, or ‘^’ in a bracket expression, put a ‘\’ in
front of it.
The exceptional handling of ] at the beginning of the list is explicitly mentioned, but to my knowledge nothing extra is said about the ^.

So this might be technically correct code, but the contract is violated.  Which puts the code at risk, if other flavours of awk are used, or if future developments change the usage of the ^.

tools/translate.sh
wlammen: Let me first illustrate how I had written the code:
```
if (stdformattedCmds[$1]) {
      # $0 pre (mm): (mm) <command> <file modified> optional <other args>
      # $0 post (bash): toolspath/inplace.sh <file modified> <command>  <other args>
      inoutfile = $2;
      $2 = cmdAsBash;
      $1 = toolspath "/inplace.sh " inoutfile;
}
```
1. You can easily verify the code does what it is expected to do.  The transformation is complete, no post-processsing necessary, nothing to be shared with other cases.
2. The contract of this piece of code (not that of the whole program) is explicitly given.  Given this is done with the other cases, too, one can verify the divide in fact sums up to the conquer (a common post-condition).
3. The variable names reflect their semantics (more) correctly.

I do not expect this exactly in this form, this has a bit of a showcase already.  But (to my liking) the support for a casual reader is too sparse in the current form.  That my premature interpretation lead to the above mentioned failure is an indication of that.  Of course it was still my fault, I should have been more careful; this could have spared me an extra round in testing, avoiding scratching my head while guessing at what went wrong.

Couldn't I have looked simply at the code, and read from it what is going on?  It's just four lines...  No.  I am not confident, no error crept in, particularly not when reviewing.

This leads to the general question, what quality standard shall we follow in future.  From the alleged scrutiny applied to the Webb telescope down to "works for me most of the time" is a wide range.  I am not in the position to set the standard here.  We shall definitely talk about how to improve review sessions, what was good/bad (https://google.github.io/eng-practices/review/reviewer/standard.html).  For now I follow the path I am used in my company, where I occasionally review other people's code.  Actually I am less rigorous here.  This is due to my lack of familiarity with the code base in total, and the techniques applied in the sources I read.

digama0: Okay. Personally I'd like to move on from this PR, so feel free to make any changes you like. This is throw-away code, so I don't think a high quality standard is necessitated in this instance. ("Works on `min.cmd` and the 3 or 4 other existing tools scripts is good enough for me.) I'm much more concerned with getting the C code up to scratch, and right now I'm bottlenecked on review. (#36 should be easy, and #37 is big but I did my best to make it straightforward.)

tools/copy.sh
```diff
@@ -0,0 +1,24 @@
...
+# Mario Carneiro 7-Jan-2022
+
+usage() {
+  cat << HELP
+COPY - Similar to Unix "cat" but safe (same input & output name allowed)
+
+This command copies (concatenates) all input files in a space separated
+list to an output file.  The output file may have
+the same name as an input file.  Any previous version of the output
+file is renamed with a ~1 extension.
```
wlammen, line 12: The backup ~1 seems to have been forgotton.

wlammen (changes requested):

tools/copy.sh
digama0: I'm just going to delete that line unless you have a good reason not to. Is that something to rely on in automated scripts?

wlammen: The automatic backup is a convenience feature for those using the tools manually and guards against unwanted data loss.  I do not care, this modification should be noted  somewhere though

benjub: Is it easy to implement the safeRename and have an option --noSafeRename that can be used by scripts ?

tools/translate.sh
```diff
@@ -0,0 +1,122 @@
...
+      $0 = file " " out;
+    } else if ($1 == "CLEAN") {
+      gsub(",", " ", $3);
+      infile = $2;
+      $2 = file;
+      $1 = tools "/inplace.sh " infile;
+    } else if ($1 == "HELP" || $1 == "COUNT" || $1 == "UPDATE") {
+      $0 = "# " $1 " command not supported";
+    } else {
+      gsub(/"/, "\\\\&");
```
wlammen, line 111: this should escape all exceptional characters (see line 116) for security reasons

wlammen (changes requested):

tools/copy.sh
digama0: moderately easy, but I have lost motivation to keep adding epicycles on this PR.

tools/translate.sh
digama0: I don't think it matters for anything other than newline, because this string is appearing in a line comment.

tools/translate.sh
```diff
@@ -0,0 +1,122 @@
...
+    }
+    if (echo) {
+      echoLine = $0;
+      gsub(/[\\$"`]/, "\\\\&", echoLine);
+      print "echo \" => " echoLine "\"";
+    }
+    print;
+    fflush();
+  }
+}'
```
wlammen, line 122: Review completed

tools/submit.sh
```diff
@@ -0,0 +1,33 @@
...
+}
+
+if [ $# -eq 1 ] && [ "$1" = "-h" -o "$1" = "--help" ]; then usage; exit; fi
+
+if [ "$1" = "-q" ]; then args=""; shift; else args="-e"; fi
+if [ $# -eq 0 ]; then tools="mm-tools"
+elif [ $# -eq 1 ]; then tools=$1
+else usage; exit 1; fi
+
+"$tools/translate.sh" -r $args "$tools" | bash
```
wlammen, line 33: Review of this file ok, and this finishes my review of this PR

PR #35, opened 2022-01-07
fix verify markup polarity

digama0:
This is a continuation of #30, part of the change to metamath.c got lost in the shuffle.


benjub:
Minor point: I see that there are many slightly different versions of the same comment (see below). Are these really the files indicated which are checked, or is it always the same set of files ? Can we give that set more explicitly/completely?
```
        1); /* 1 = check missing external files (gifs, bib, etc.) */

    flag fileCheck, /* 1 = check external files (gifs, mmset.html,...) */

          fileCheck); /* 1 = check missing external files (gifs, bib, etc.) */

/* If checkFiles = 0, do not open external files.
   If checkFiles = 1, check mm*.html, presence of gifs, etc. */

H("    / FILE_CHECK - This qualifier will include checks that require");
H("        external files to be present, such as checking GIF existence and");
H("        bibliographic links to mmset.html or equivalent.");

   flag fileCheck)  /* 1 = check for missing external files (gifs, bib, etc.) */

   flag fileCheck) /* 1 = check missing external files (mmbiblio.html) */

    flag fileCheck); /* 1 = check missing external files (gifs, bib, etc.) */
```
Also, the comment in the line `flag topDateCheck, /* 1 = check top date but check others */` sounds strange (maybe remove "but check others"?)

digama0:
From what I can tell, the "external files" check checks the following:

* For each `htmldef` definition containing `... IMG SRC="bla.gif" ...`, `bla.gif` must exist or a warning is printed.
* The bibliography file provided by the `htmlbibliography` or `exthtmlbibliography` command must exist or a warning is printed.
* The list of all `<A NAME="foo"` links in the bibliography page is collected, and if an anchor is duplicated then an error is printed.
* If the bibliography has been loaded (due to setting `fileCheck` in this or a previous call), then any reference to a tag that does not exist in the bibliography causes an error to be printed.

The abbreviated comments are thus broadly correct. Note that sometimes it says `bib` and sometimes `mmset.html`, because the bibliography file for set.mm is `mmset.html`. I don't think `mmbiblio.html` is correct: this is an output of compilation, not an input, so metamath never does an existence check on this file except when replacing it.

benjub:
Thanks for the explanation.  This should be kept somewhere.  Maybe in a comment in the relevant .c file ?

Still, it would be nice to have more uniform comments, if you find a good phrasing.

wlammen (approved):
No issues found

benjub:
Could the information written above by @digama0 find its way somewhere in the documentation (probably in the relevant .c file) ?

digama0:
I already added it in a4fd612

PR #36, opened 2022-01-07
reduce imports

digama0:


mmvstr.h
```diff
@@ -124,6 +124,8 @@ with 'let(&' and thus has the same effect as 'let(&'.
 #ifndef METAMATH_MMVSTR_H_
 #define METAMATH_MMVSTR_H_
 
+#include <stdio.h>
```
digama0, line 127: This was a required include that was masked by the fact that `#include <stdio.h>` always appears before `#include "mmvstr.h"` in every c file that uses the latter.

jkingdon (approved):

PR #37, opened 2022-01-07
add types for temp_(v/nmbr/pntr)String, use const

digama0:
The memory management pays a lot of attention to whether a returned `vstring` is to be deallocated by the caller, or it is a temporary value which is automatically deallocated by the next call to `let`. This commit makes it clear which is which by using the return type of the function (where `temp_vstring` is a typedef for `vstring`.)

Similarly, a `vstring` is not actually the same as a C `char *`: a `vstring` can be viewed as a `char *`, but it also has a control block at negative addresses which can be used to reallocate the string. Most functions that take `const vstring` are in fact `const char *` methods, so the type signatures have been changed to reflect this. Similarly, many input arguments can be `const`, which is a C99 thing (which was already in the codebase), so use it more consistently.

(This is one part of a big commit that I'm breaking into pieces for ease of review.)

### DIFF
--- a/src/mmdata.c
+++ b/src/mmdata.c
@@ -56,15 +56,15 @@ struct nullNmbrStruct g_NmbrNull = {-1, sizeof(long), sizeof(long), -1};
 /* Null ptrString */
 struct nullPntrStruct g_PntrNull = {-1, sizeof(long), sizeof(long), NULL};
 
-nmbrString *nmbrTempAlloc(long size);
+temp_nmbrString *nmbrTempAlloc(long size);
         /* nmbrString memory allocation/deallocation */
-void nmbrCpy(nmbrString *sout, nmbrString *sin);
-void nmbrNCpy(nmbrString *s, nmbrString *t, long n);
+void nmbrCpy(nmbrString *sout, const nmbrString *sin);
+void nmbrNCpy(nmbrString *s, const nmbrString *t, long n);
 
-pntrString *pntrTempAlloc(long size);
+temp_pntrString *pntrTempAlloc(long size);
         /* pntrString memory allocation/deallocation */
-void pntrCpy(pntrString *sout, pntrString *sin);
-void pntrNCpy(pntrString *s, pntrString *t, long n);
+void pntrCpy(pntrString *sout, const pntrString *sin);
+void pntrNCpy(pntrString *s, const pntrString *t, long n);
 
 vstring g_qsortKey; /* Used by qsortStringCmp; pointer only, do not deallocate */
 
@@ -433,8 +433,7 @@ long getFreeSpace(long max)
 }
 
 /* Fatal memory allocation error */
-void outOfMemory(vstring msg)
-{
+void outOfMemory(const char *msg) {
   vstring tmpStr = "";
   print2("*** FATAL ERROR:  Out of memory.\n");
   print2("Internal identifier (for technical support):  %s\n",msg);
@@ -558,7 +557,7 @@ void bug(int bugNum)
 
 /* This function returns a 1 if any entry in a comma-separated list
    matches using the matches() function. */
-flag matchesList(vstring testString, vstring pattern, char wildCard,
+flag matchesList(const char *testString, const char *pattern, char wildCard,
     char oneCharWildCard) {
   long entries, i;
   flag matchVal = 0;
@@ -588,7 +587,7 @@ flag matchesList(vstring testString, vstring pattern, char wildCard,
    the second argument.  The second argument may have wildcard characters.
    wildCard matches 0 or more characters; oneCharWildCard matches any
    single character. */
-flag matches(vstring testString, vstring pattern, char wildCard,
+flag matches(const char *testString, const char *pattern, char wildCard,
     char oneCharWildCard) {
   long i, ppos, pctr, tpos, s1, s2, s3;
   vstring tmpStr = "";
@@ -730,10 +729,10 @@ long g_nmbrTempAllocStackTop = 0;     /* Top of stack for nmbrTempAlloc functon 
 long g_nmbrStartTempAllocStack = 0;   /* Where to start freeing temporary allocation
                                     when nmbrLet() is called (normally 0, except in
                                     special nested vstring functions) */
-nmbrString *nmbrTempAllocStack[M_MAX_ALLOC_STACK];
+temp_nmbrString *nmbrTempAllocStack[M_MAX_ALLOC_STACK];
 
 
-nmbrString *nmbrTempAlloc(long size)
+temp_nmbrString *nmbrTempAlloc(long size)
                                 /* nmbrString memory allocation/deallocation */
 {
   /* When "size" is >0, "size" instances of nmbrString are allocated. */
@@ -788,7 +787,7 @@ void nmbrMakeTempAlloc(nmbrString *s)
 /* to work properly.  If a nmbrString has never been assigned before, */
 /* it is the user's responsibility to initialize it to NULL_NMBRSTRING (the */
 /* null string). */
-void nmbrLet(nmbrString **target,nmbrString *source) {
+void nmbrLet(nmbrString **target, const nmbrString *source) {
   long targetLength,sourceLength;
   long targetAllocLen;
   long poolDiff;
@@ -896,17 +895,14 @@ void nmbrLet(nmbrString **target,nmbrString *source) {
 
 
 
-nmbrString *nmbrCat(nmbrString *string1,...) /* String concatenation */
+temp_nmbrString *nmbrCat(const nmbrString *string1,...) /* String concatenation */
 #define M_MAX_CAT_ARGS 30
 {
   va_list ap;   /* Declare list incrementer */
-  nmbrString *arg[M_MAX_CAT_ARGS];        /* Array to store arguments */
+  const nmbrString *arg[M_MAX_CAT_ARGS];        /* Array to store arguments */
   long argLength[M_MAX_CAT_ARGS];       /* Array to store argument lengths */
-  int numArgs=1;        /* Define "last argument" */
-  int i;
-  long j;
-  nmbrString *ptr;
-  arg[0]=string1;       /* First argument */
+  int numArgs = 1;        /* Define "last argument" */
+  arg[0] = string1;       /* First argument */
 
   va_start(ap,string1); /* Begin the session */
   while ((arg[numArgs++]=va_arg(ap,nmbrString *)))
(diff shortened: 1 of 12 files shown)


wlammen:
Yep.  Necessary.  I already thought of similar measures.

mmvstr.h
```diff
@@ -126,18 +126,26 @@ with 'let(&' and thus has the same effect as 'let(&'.
```
wlammen, line 126: I'd like to see a usage example for temp_vstring in the introducing comment block.  Examples should be updated to use vstringdef.

mmvstr.c
```diff
@@ -91,8 +91,7 @@ static void* tempAlloc(long size)  /* String memory allocation/deallocation */
 /* Make string have temporary allocation to be released by next let() */
```
wlammen, line 91: How is the connection between makeTempAlloc and temp_vstring?

wlammen (changes requested):

mmvstr.c
```diff
@@ -138,25 +137,23 @@ void let(vstring *target, vstring source) {
...
     }
-    *target= "";
+    *target = "";
   }
 
   freeTempAlloc(); /* Free up temporary strings used in expression computation */
 
 } /* let */
 
-vstring cat(vstring string1,...)        /* String concatenation */
```
wlammen, line 148: cat has another limitation not mentioned in the comments of master: the list of sources must not be empty i.e cat(NULL) is illegal

wlammen (changes requested):

mmvstr.c
```diff
@@ -234,50 +225,44 @@ int linput(FILE *stream, const char* ask, vstring *target)
 
 
 /* Find out the length of a string */
-long len(vstring s)
-{
-  return ((long)strlen(s));
+long len(vstring s) {
```
wlammen, line 228: is there a particular reason why len does not take a char const* as parameter?

mmvstr.c
```diff
@@ -669,15 +645,13 @@ vstring xlate(vstring sin,vstring table)
 
 
 /* Returns the ascii value of a character */
-long ascii_(vstring c)
-{
+long ascii_(const char *c) {
   return ((long)c[0]);
```
wlammen, line 649: The implementation hazardous, because for a non-ASCII character the result depends on the implementation defined signedness of char (signed or unsigned char, you cannot tell).  This is not something that has to do with your PR, but it caught my eye.

wlammen (changes requested):

mmdata.c
```diff
@@ -923,39 +919,39 @@ nmbrString *nmbrCat(nmbrString *string1,...) /* String concatenation */
...
 
 
 
 /* Find out the length of a nmbrString */
-long nmbrLen(nmbrString *s)
+long nmbrLen(const nmbrString *s)
 {
   /* Assume it's been allocated with poolMalloc. */
-  return (((long)(((long *)s)[-1] - (long)(sizeof(nmbrString))))
+  return (((long)(((const long *)s)[-1] - (long)(sizeof(nmbrString))))
```
wlammen, line 945: I don't think, the cast to long of the nominator is necessary:
((((const long *)s)[-1] - (long)(sizeof(nmbrString)))
is sufficient and still looks complex.
same holds for line 954

mmdata.c
```diff
@@ -1137,8 +1118,7 @@ long nmbrRevInstr(long start_position, nmbrString *string1,
 
 
 /* Converts nmbrString to a vstring with one space between tokens */
-vstring nmbrCvtMToVString(nmbrString *s)
-{
+temp_vstring nmbrCvtMToVString(const nmbrString *s) {
   long i, j, outputLen, mstrLen;
   vstring tmpStr = "";
```
wlammen, line 1123: Why do you avoid use of vstringdef here?
Suggest to move the declaration of ptr and ptr2 right before the for loop.
The implementation looks spooky if s is empty. In this case space seems to alloc an entry on the tempAllocStack containing the empty string, but where is that freed? let() ignores freeing of empty strings (?).  There seems to be a memory leak.

wlammen (changes requested):

mmdata.c
digama0: I'd rather not mess with uses of `vstringdef`, because it will conflict with #38 which does a bulk translation to use the macro everywhere (and it also changes the name from `vstringdef` to `vstring_def` so added references to `vstringdef` in this PR will also cause conflicts). I think if you look you will likely find that this line is touched in #38.

digama0: Same thing about the other suggestions: this PR is about constifying and nothing more (to the extent possible). I have a bunch more changes waiting in the wings but if I put it all in one PR it will be huge and hard to follow, so I'm trying to do one type of change at a time.

digama0: Based on how the code generally handles these temp strings, it seems like empty strings never have to be deallocated. I assume the reason is because `let` will never create an empty string, it always sets the input to some designated `""` in static memory. (If that's not the case, then the whole `let(&x, "")` idiom is suspect, but even beyond that there are parts of the code that say that skip the `let(&x, "")` statement if `!x[0]`, which is bogus if "non-canonical" empty strings can be created in temporary storage.

digama0: Canonicalization of empty strings is done here:

https://github.com/metamath/metamath-exe/blob/ddcd53aa5b6e9403f3692ebde8b42f0dc24aff7f/mmvstr.c#L128-L134

digama0: In this function `nmbrCvtMToVString`, when `s` is empty, it does:
* `mstrLen = 0`
* `outputLen = -1`
* `let(&tmpStr, "")` (so `tempStr` is a pointer to `""` in static storage)
* The `if (tmpStr[0]) makeTempAlloc(tmpStr);` line is skipped
* `return tmpStr`

So it returns a canonical empty `vstring`, which is fine to be treated as a `temp_vstring` because it doesn't need to be deallocated.

mmdata.c
digama0: There are a lot of redundant casts like this. I think it was being compiled in some mode where the compiler would give a warning if you did not use a cast even for implicit numeric up-casts, but I don't think our current compiler settings enforce this. I think I want to clean this up in a separate PR. (The cast `const long *` from `const nmbrString *` is unnecessary since `nmbrString` is a typedef for `long`.) The casts here look like defense against promotion in the division; probably if we read the C spec it can be simplified a lot.

mmdata.c
wlammen: Under the assumption outputLen == -1
in mmdata.c line 1137 nmbrCvtMToVStr calls space(-1)
in mmvstr.c line 571 space(-1) calls string(-1, ' ')
in mmvstr.c line 562 string(0, ' ') calls tempAlloc(1) 
In mmvstr.c line 56: tempAlloc(1) allocates 1 byte memory
Question: when is this byte freed again?
This is important, since you strengthen the return type from vstring to temp_vstring.

digama0: Aha, this is a bug in `string()`, it should return `""` if `n = 0`

wlammen: One should check whether nmbrSeg e. a. suffer from a similar problem.

digama0:
I'd like to get this merged before #38. Are there any more objections? The `string()` bug was not introduced in this PR and will be handled in its own PR.

mmdata.c
digama0: On further review (while preparing a separate PR for this issue), I think there is actually no bug here. The invariant is as follows:

* Temporary strings (`temp_vstring`, constructed by `tempAlloc` and freed by `freeTempAlloc`, usually via `let`) are allowed to be noncanonical empty strings, and are also allowed to have internal `\0` characters as a quick truncation operation. They should never be manually freed, because there is no easy means to remove them from the temp stack, and this is why they cannot appear in the left argument of a `let`. Static strings can also be treated as `temp_vstring`, because from client code this only means that the string should not be freed manually and one cannot assume that they will live past the next call to `freeTempAlloc`.

* Regular strings (`vstring`, constructed by `vstring_def`) can be in one of two states: empty, and pointing to the canonical empty string `""`, or nonempty and pointing to a `malloc`'d buffer with a control block at bytes -12..0 relative to the pointer. The `let(&str, value)` function can be used to change the value of a `vstring str`. `vstring` variables always start in the empty state using `vstring_def`, and they must be in the empty state when exiting a scope or else memory will leak.

As it applies to this code, we see that the invariants are upheld:

* `string` has no bug: it may sometimes create noncanonical empty strings but they are always temp-allocated.
* `nmbrCvtMToVString` uses `let(&tmpStr, space(outputLen))` correctly to copy a `temp_vstring` into a `vstring`, and calls `makeTempAlloc` to put it on the temporary stack. It will be deallocated at the next call to `let`.
* `makeTempAlloc` doesn't have a bug per se, but it has IMO the wrong precondition, and one that does not match the corresponding functions `nmbrMakeTempAlloc` and `pntrMakeTempAlloc`. It is supposed to convert a `vstring` into a `temp_vstring` and the way to do that is to put owned (nonempty) `vstring`s on the temp stack for later reclamation, and do nothing for (canonical static) empty `vstring`s since this already qualifies as a `temp_vstring`, it should not be deallocated. Currently, the nonempty check is done outside the function for `makeTempAlloc` and inside the function for `nmbrMakeTempAlloc` and `pntrMakeTempAlloc`; I will fix this in another PR.

src/mmdata.c
```diff
@@ -1137,8 +1118,7 @@ long nmbrRevInstr(long start_position, nmbrString *string1,
 
 
 /* Converts nmbrString to a vstring with one space between tokens */
-vstring nmbrCvtMToVString(nmbrString *s)
-{
+temp_vstring nmbrCvtMToVString(const nmbrString *s) {
   long i, j, outputLen, mstrLen;
   vstring tmpStr = "";
```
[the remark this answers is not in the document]
digama0: It is born as a `vstring`, and is only converted to a `temp_vstring` right at the end. I think type ascribing it as a `temp_vstring` from the beginning would be confusing, but there is no real way in C to express a function that changes the type of its argument. How about having `makeTempString` return its argument instead (but with a different type)?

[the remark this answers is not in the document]
digama0: Conceptually, you should not think of it as ever bypassing the `makeTempString` call. It is always converted to a temp, it is just that sometimes there is nothing to be done. #46 makes this clearer.

src/mmvstr.h
```diff
@@ -126,18 +126,69 @@ with 'let(&' and thus has the same effect as 'let(&'.
 
 #include <stdio.h>
 
+/* A vstring is like a C string, but it contains a control block allowing
+for memory allocation. New vstrings should always be constructed from the
+`vstringdef` macro. */
 typedef char* vstring;
+
+/* A vstring allocated in temporary storage. These strings will be deallocated
```
wlammen, line 134: ...vstring is allocated...

src/mmdata.c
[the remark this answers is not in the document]
wlammen: > How about having makeTempString return its argument instead (but with a different type)?
That's cool.

wlammen (changes requested):

src/mmdata.c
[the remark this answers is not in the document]
wlammen: Fix the typo, Nothing more.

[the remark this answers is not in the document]
digama0: what typo? Are we still talking about this line or about #46?

src/mmvstr.h
digama0:
```suggestion
/* A vstring is allocated in temporary storage. These strings will be deallocated
```

PR #38, opened 2022-01-09
add macros for alloc/free of string types

digama0:
* This commit adds macros `free_vstring`, `free_nmbrString` and `free_pntrString` for the idiom `let(&x, "")`, in order to make clear when this is being used to deallocate memory (as opposed to actually setting the value to the empty string).
* It also adds macros `vstring_def`, `nmbrString_def` and `pntrString_def` for the idiom `vstring x = ""`, because this is the only safe way to construct a new `vstring`. (The macro `vstringdef` already existed; it has been renamed to `vstring_def` for consistency with the others and used throughout.)
* In a few places, the idiom `let(&dummy, "")` was being used for its side effect of clearing temporary variables, for example when `mid()` is being called in a loop which does not otherwise contain a `let`. Instead of this, the function `freeTempAlloc()` that actually does the work has been made public so it can be called directly.
* Other than whitespace and parenthesization changes, there is nothing else in this commit, despite its large number of touched lines. It is essentially a search and replace job.

### DIFF
--- a/src/metamath.c
+++ b/src/metamath.c
@@ -758,13 +758,17 @@ void command(int argc, char *argv[]) {
   long stmt, step;
   int subType = 0;
 #define SYNTAX 4
-  vstring str1 = "", str2 = "", str3 = "", str4 = "", str5= "";
+  vstring_def(str1);
+  vstring_def(str2);
+  vstring_def(str3);
+  vstring_def(str4);
+  vstring_def(str5);
   nmbrString *nmbrTmpPtr; /* Pointer only; not allocated directly */
-  nmbrString *nmbrTmp = NULL_NMBRSTRING;
-  nmbrString *nmbrSaveProof = NULL_NMBRSTRING;
+  nmbrString_def(nmbrTmp);
+  nmbrString_def(nmbrSaveProof);
   /*pntrString *pntrTmpPtr;*/ /* Pointer only; not allocated directly */
-  pntrString *pntrTmp = NULL_PNTRSTRING;
-  pntrString *expandedProof = NULL_PNTRSTRING;
+  pntrString_def(pntrTmp);
+  pntrString_def(expandedProof);
   flag tmpFlag;
 
   /* proofSavedFlag tells us there was at least one
@@ -794,21 +798,21 @@ void command(int argc, char *argv[]) {
   flag saveFlag; /* Flag to save in source */
   flag fastFlag; /* Flag for SAVE PROOF.../FAST */
   long indentation; /* Number of spaces to indent proof */
-  vstring labelMatch = ""; /* SHOW PROOF <label> argument */
+  vstring_def(labelMatch); /* SHOW PROOF <label> argument */
 
   flag axiomFlag; /* For SHOW TRACE_BACK */
   flag treeFlag; /* For SHOW TRACE_BACK */
   flag countStepsFlag; /* For SHOW TRACE_BACK */
   flag matchFlag; /* For SHOW TRACE_BACK */
-  vstring matchList = "";  /* For SHOW TRACE_BACK */
-  vstring traceToList = ""; /* For SHOW TRACE_BACK */
+  vstring_def(matchList);  /* For SHOW TRACE_BACK */
+  vstring_def(traceToList); /* For SHOW TRACE_BACK */
   flag recursiveFlag; /* For SHOW USAGE */
   long fromLine, toLine; /* For TYPE, SEARCH */
   flag joinFlag; /* For SEARCH */
   long searchWindow; /* For SEARCH */
   FILE *type_fp; /* For TYPE, SEARCH */
   long maxEssential; /* For MATCH */
-  nmbrString *essentialFlags = NULL_NMBRSTRING;
+  nmbrString_def(essentialFlags);
                                             /* For ASSIGN/IMPROVE FIRST/LAST */
   long improveDepth; /* For IMPROVE */
   flag searchAlg; /* For IMPROVE */
@@ -821,7 +825,7 @@ void command(int argc, char *argv[]) {
   flag commentOnlyFlag; /* For SHOW STATEMENT */
   flag briefFlag; /* For SHOW STATEMENT */
   flag linearFlag; /* For SHOW LABELS */
-  vstring bgcolor = ""; /* For SHOW STATEMENT definition list */
+  vstring_def(bgcolor); /* For SHOW STATEMENT definition list */
 
   flag verboseMode, mayGrowFlag /*, noDistinctFlag*/; /* For MINIMIZE_WITH */
   long prntStatus; /* For MINIMIZE_WITH */
@@ -831,13 +835,13 @@ void command(int argc, char *argv[]) {
   long thisMathboxStartStmt; /* For MINIMIZE_WITH */
   flag forwFlag; /* For MINIMIZE_WITH */
   long forbidMatchPos;  /* For MINIMIZE_WITH */
-  vstring forbidMatchList = "";  /* For MINIMIZE_WITH */
+  vstring_def(forbidMatchList);  /* For MINIMIZE_WITH */
   long noNewAxiomsMatchPos;  /* For NO_NEW_AXIOMS_FROM */
-  vstring noNewAxiomsMatchList = "";  /* For NO_NEW_AXIOMS_FROM */
+  vstring_def(noNewAxiomsMatchList);  /* For NO_NEW_AXIOMS_FROM */
   long allowNewAxiomsMatchPos;  /* For NO_NEW_AXIOMS_FROM */
-  vstring allowNewAxiomsMatchList = "";  /* For NO_NEW_AXIOMS_FROM */
-  vstring traceProofFlags = ""; /* For NO_NEW_AXIOMS_FROM */
-  vstring traceTrialFlags = ""; /* For NO_NEW_AXIOMS_FROM */
+  vstring_def(allowNewAxiomsMatchList);  /* For NO_NEW_AXIOMS_FROM */
+  vstring_def(traceProofFlags); /* For NO_NEW_AXIOMS_FROM */
+  vstring_def(traceTrialFlags); /* For NO_NEW_AXIOMS_FROM */
   flag overrideFlag; /* For discouraged statement /OVERRIDE */
 
   struct pip_struct saveProofForReverting = {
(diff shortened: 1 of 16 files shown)


mmcmds.c
```diff
@@ -4936,9 +4937,9 @@ void verifyMarkup(vstring labelMatch,
       continue;
     }
 
-    let(&str1, ""); /* Prevent string stack buildup/overflow in left() below */
     /* Look for "ax-*" axioms */
     if (strcmp("ax-", left(g_Statement[stmtNum].labelName, 3))) {
+      freeTempAlloc(); /* Prevent string stack buildup/overflow in left() */
```
wlammen, line 4942: What is going on here?

mmdata.c
```diff
@@ -1121,26 +1104,24 @@ long nmbrRevInstr(long start_position, nmbrString *string1,
...
-   nmbrString *tmp = NULL_NMBRSTRING;
    ls1 = nmbrLen(string1);
    ls2 = nmbrLen(string2);
    if (start_position > ls1 - ls2 + 1) start_position = ls1 - ls2 + 2;
    if (start_position<1) return 0;
    while (!nmbrEq(string2, nmbrMid(string1, start_position, ls2))) {
      start_position--;
-     nmbrLet(&tmp, NULL_NMBRSTRING);
-              /* Clear nmbrString buffer to prevent overflow caused by "mid" */
+     nmbrTempAlloc(0); /* Clear temporaries to prevent overflow caused by "mid" */
```
wlammen, line 1113: Dont know whether this is correct.

mmdata.c
```diff
@@ -2683,15 +2664,14 @@ long pntrRevInstr(long start_position,pntrString *string1,
...
-   pntrString *tmp = NULL_PNTRSTRING;
    ls1=pntrLen(string1);
    ls2=pntrLen(string2);
    if (start_position>ls1-ls2+1) start_position=ls1-ls2+2;
    if (start_position<1) return 0;
    while (!pntrEq(string2,pntrMid(string1,start_position,ls2))) {
      start_position--;
-     pntrLet(&tmp,NULL_PNTRSTRING);
-                /* Clear pntrString buffer to prevent overflow caused by "mid" */
+     pntrTempAlloc(0);
```
wlammen, line 2673: Again: change this in a different PR

mmvstr.c
```diff
@@ -43,8 +43,7 @@ long g_startTempAllocStack = 0;    /* Where to start freeing temporary allocatio
                                     special nested vstring functions) */
 void *tempAllocStack[MAX_ALLOC_STACK];
 
-static void freeTempAlloc(void)
-{
+void freeTempAlloc(void) {
```
wlammen, line 46: This changes visibility during linking

mmvstr.h
```diff
@@ -106,14 +106,13 @@ outside of 'let(&' assignments; for example
...
+If necessary, 'freeTempAlloc' can be used periodically to clear
 this temporary memory:
 
-        for (i=0; i<10000; i++)
-          {
+        for (i=0; i<10000; i++) {
           print2("%s\n",left(string1,70));
-          let(&dummy,"");
-          }
+          freeTempAlloc();
```
wlammen, line 114: This seems to be part of a comment.  Is this semantically correct?

mmvstr.h
```diff
@@ -127,7 +126,13 @@ with 'let(&' and thus has the same effect as 'let(&'.
...
 
 typedef char* vstring;
-#define vstringdef(x) vstring x = ""
+#define vstring_def(x) vstring x = ""
+#define free_vstring(x) let(&x, "")
+
+/* Free space allocated for temporaries. This is normally called automatically
+  by let(), but it can also be called directly to avoid buildup of temporary
+  strings. */
+void freeTempAlloc(void);
```
wlammen, line 135: This affects other artefacts as well.  Difficult to see whether this is correct.

wlammen (changes requested):
This is a massive change.  Occasionally something has been fixed "while I am at it".  It is difficult for me to decide whether that is correct.  I suggest to do the bulk in this PR and do the fine tuning in a different one

mmcmds.c
digama0: The call to `left` produces a temporary. Usually such temporaries are used in an argument to `let` so they are cleaned up thereafter, but in this case it is an argument to `strcmp` which does no such cleanup. This is not normally an issue since there are `let`s all over the place, but in this particular case there is the possibility that we get to L4941, allocate the temporary, and then `continue` at L4943, leaving the temp uncleared. If this happens for many loop iterations in a row then we can overflow the temp buffer.

The issue only happens in this branch: on previous `continue`s in this loop no temporary has been created, and after this if statement there is a `let` which will clean up the temporary.

The `let(&str1, "")` command is otherwise useless; `str1` is not in use at this point in the program. (I find it clearer to not randomly set variables that aren't in use.)

mmdata.c
digama0: This pattern of clearing an already cleared variable was the previous idiom for calling `let` for its side effect of clearing the temp buffer. I think that it is better not to involve a dead variable and confuse the reader. In this case it is pretty easy to see that `tmp` is otherwise useless: it is only ever set to `NULL_NMBRSTRING`.

mmvstr.c
digama0: Yes? It's public now

mmvstr.h
digama0: Yes, this is updating the comment to stop recommending the idiom of using a dummy and call `freeTempAlloc` directly.

digama0:
The `freeTempAlloc` change is in this PR because it is directly related. The goal here was to classify every existing use of `let(&var, "")` into three cases:

* `var` is actually being set to `""`, semantically, for example because we are going to concatenate to it. In this case, we keep it as `let(&var, "")`.
* `var` is being deallocated, either because it will be reassigned using `var = ...` from a function that returns an owned `vstring`, or because we are exiting a scope.  In this case we use the new macro `free_vstring(var)`.
* `var` is a dummy variable which is either already cleared or introduced specifically for the purpose of this statement, and `let()` is being used for its side effect of clearing the temporary stack. In this case we use `freeTempAlloc()`.

mmdata.c
digama0: Addressed on PR comment

mmcmds.c
wlammen: The problem is that freeTempAlloc not only frees the temporary created by left, but other temporaries too.  Can you tell without thinking that str2, defined in line 4852:
let(&str2, cat(",", edit(g_Statement[stmtNum].labelName, 32/*uppercase*/),
        ",", NULL));
is not reused later?
While str2 is redefined here short after (so this particular case does not hurt), I had to make sure no variables are used ACROSS this loop.  And how do you know someone adds one in future, not knowing it is destroyed here?  This technique is inherently unsafe.

At least put a 'memory barrier' in form of a comment:
/* do not use temporary variables across this line */
in front of this loop.  Then there is at least an assertion to fulfil.

mmdata.c
wlammen: Again: this frees temporaries allocated by callers of this function, too.  How is it guaranteed this never occurs now and in future?

mmdata.c
wlammen: See my previous remark.  As a bare minimum you must formulate a precondition no caller has allocated something on the stack that it wants to reuse after the call to this function.

mmvstr.c
wlammen: Very unlikely, but it may distract previous calls to some other freeTempAlloc defined elsewhere.  You have to verify this detail, and I wasn't inclined to do so any more after having evaluated several thousand lines of code.  See my introductory note.

grep shows this is not the case.  So accepted.

mmvstr.h
wlammen: The risk inherent to freeTempAlloc should be noted somewhere

mmvstr.h
wlammen: Forget about this one.  Possible collisions are mentioned in one of my earlier remarks

wlammen (changes requested):

mmdata.c
digama0: Every call to `let` has this same effect, so if the code is broken now it was already broken before. Except in very specialized circumstances (specifically, inside the implementation of some of the temp-producing functions), the temp stack is always completely cleared after every `let`, so the lifetime of temporaries in most of the code is syntactically evident (from the temp-producing call to the next call to `let` or `freeTempAlloc`).

mmcmds.c
digama0: > The problem is that freeTempAlloc not only frees the temporaries created by left, but other temporaries too. Can you tell without thinking that str2, defined in line 4852: let(&str2, cat(",", edit(g_Statement[stmtNum].labelName, 32/_uppercase_/), ",", NULL)); is not reused later?

Yes, I can tell this easily. `str2` is not a temporary (a `temp_vstring`), it is a regular `vstring`. Temporaries *must not* appear on the left of a `let` statement, and they must be cleaned up at the end of the scope (usually, the end of the function). They are not subject to cleanup by `let` statements.

A temporary would be something like the return value of `edit()` in that example. These are almost never stored in variables, they only exist as subexpressions (C temporaries), and they usually only last for the duration of one line of code, so it is generally pretty easy to see their scope.

As I mention in the other comment, `g_startTempAllocStack` is almost always 0, except inside the implementation of some vstring functions. There are no "local temporary scopes" in the main part of the code.

mmdata.c
digama0: This is documented here:

https://github.com/metamath/metamath-exe/blob/8b442fa82600ea7892417fe0da17acd12fc3fbf8/src/mmvstr.c#L41-L43

digama0: Here's another example that shows that these sections are handled with care:

https://github.com/metamath/metamath-exe/blob/8b442fa82600ea7892417fe0da17acd12fc3fbf8/src/mminou.c#L386-L398

mmdata.c
digama0: Note that `nmbrTempAlloc(0)` appears literally at the end of `nmbrLet()`. This does not reset the stack to 0, it resets the stack to `g_nmbrStartTempAllocStack`, so this will actually work even if `g_nmbrStartTempAllocStack` is nonzero, although as I mention in another comment this only happens in exceptional situations, and most functions have the precondition that `g_nmbrStartTempAllocStack` is zero.

wlammen (approved):
OK, go on then.

digama0:
Just checking, did you mean to respond to https://github.com/metamath/metamath-exe/pull/37#issuecomment-1022604689 ? Or are you good with the contents of this PR as well?

PR #44, opened 2022-01-24
move sources to specific folder

wlammen:
Completes item 'Move source code to a src/ directory' in #17.

Rationale: Before I can add a Doxyfile for generation of documentation I need to know where the source folder is.  This PR moves the sources into a new folder.

@ Reviewer: Check the contents of README.TXT whether following the instructions creates an executable.  The executable is generated in the src folder.  This need not be the optimal location, in particular with regard to the tools folder.  Suggestions?

Merging this PR requires updates of pending pull requests.

The configure.ac file has issues:
AC_INIT([metamath], [0.114], [nm.NOSPAM@alum.mit.edu])
The version should be 0.199-something, and the email address for reporting bugs should be redirected to someone willing to handle these.


benjub:
In the github repositories I follow, the files Makefile.am and configure.ac are generally in the root directory.
For the executable, maybe a `bin/` directory ?

digama0:
I would want all the autoconf garbage to be generated in the `build/` folder; I'm ambivalent about whether the executable should also end up in there or get moved to the `bin/` directory when it is done.

wlammen:
...to be continued

benjub:
> I'm ambivalent about whether the executable should also end up in there or get moved to the `bin/` directory when it is done.

Even if this may sound overkill, I think having a separate `bin/` would make things clearer.

digama0:
I mean to say that conventions are somewhat split on whether to put generated binaries in `bin/` or somewhere in `build/`.

* Rust `cargo` will generate the executable in `target/release/myprog`, and having all generated files in a single directory makes it easier to do the equivalent of a manual `make clean` if needed.
* Haskell `stack` does the same thing, although the executable seems to be in an undisclosed location somewhere inside the `.stack-work` directory, and `stack install` copies the executable into a global location like `~/.local/bin`.
* I'm not super familiar with go project structure but it seems like putting the executables in the `build` directory is also common.
*  [Lean](https://github.com/leanprover-community/lean/) is an example of a (C++) project I know of that puts its executables in `bin/`. More specifically, build outputs are all generated in `build/debug/` or `build/release/` but when you use the default `make` target one of the final steps is to copy the executable into the `bin/` directory (which actually has a few checked in files as well, some shell scripts and executable python files).

As I see it the primary advantages are:
* If you put all generated files in `build/`, it is easier for the user to know what is generated and what isn't, so they can safely delete the generated files.
* If you put the executable in `bin/`, then it is easier to locate the executable after running some voodoo build command.
* A third option is to use the `make install` approach and put the executable in `.local/bin` or so, so that you can directly start running it after it is built without worrying where it was placed. But this should probably be a conscious choice for the user, they may not want random files getting installed without their consent.

wlammen:
I added a build folder with a build.sh script in it.  If you change to that folder and run the script, a metamath executable is compiled WITHOUT touching the src/ folder, except that the executable is copied there.  All build files are still kept in the build folder, for inspection.

wlammen:
I'd like to finish this PR.  The documentation can still be enlarged, and maybe the build can be improved or made more error tolerant.  But my initial incentive was to move the sources to a defined location to start Doxygen installation.

digama0:
The executable is copied into the `src/` folder? That seems unusual...

wlammen:
That's where it was created before, together with all the build files.  There is some discussion going on, whether build/ or bin/ is the better location.  This is the last line in the script.  So feel free to change that line to your liking, once it is in the sources.  I don't want to let fine tuning block this PR more than necessary.

build/build.sh
```diff
@@ -0,0 +1,45 @@
...
+
+rm $BUILDDIR/configure.ac.orig
+
+#===========   run autoconf   =====================
+
+cd $BUILDDIR
+autoreconf -i
+./configure
+make
+cp -f metamath $SRCDIR
```
digama0, line 48:
```suggestion
```
Let's not move it anywhere if we aren't sure where to put it. `/`, `build/` and `bin/` are possible locations but `src/` is definitely wrong.

PR #46, opened 2022-01-26
makeTempAlloc returns self instead of type-changing

digama0:
* Depends on #37

Fix for the issue described in https://github.com/metamath/metamath-exe/pull/37#discussion_r793080285 . The `makeTempAlloc` functions now take a `vstring` and return a `temp_vstring`, rather than mutating the value in place and effectively changing its type. In all existing situations the result was immediately returned so this results in simpler code anyway.

Note that the `makeTempAlloc(x)` call was previously guarded by a condition `x[0] != 0`, because (canonical) empty strings should not be put on the temp stack. This condition has been moved inside the function, so that it can have a clearer contract.


wlammen (approved):
OK

PR #48, opened 2022-01-27
avoid sed if possible

wlammen:
Reissued PR after network breakdown caused sync trouble between the GitHub repository and my local copy.
Fixed links.  Extended documentation.


wlammen:
No further commits here, unless sth is broken.

PR #49, opened 2022-01-28
add Doxyfile

wlammen:
The build.sh script moved to the src folder now. If doxygen is installed on your system, the documentation is built as well, in the build/html folder. Check it out. cd to metamath-exe/src and run ./build.sh. Results are in metamath-exe/build.  Run build.sh -h for some basic help.


wlammen:
The build.sh script moved to the src folder now.  If doxygen is installed on your system, the documentation is built as well, in the build/html folder.  Check it out.

digama0:
Is it legal for `Makefile.am.annotated` to actually replace `Makefile.am`? I don't like the idea of having `Makefile.am` data duplicated since the whole point of all the autotools stuff is to avoid having to put that list of files in multiple places.

digama0:
The build command is starting to look like a lot more than `autoconf && ./configure && make`. Shouldn't this stuff be part of the makefile? Also I think you aren't supposed to run `./configure` every time.

src/metamath.c
```diff
@@ -54,8 +58,21 @@
...
-
-#define MVERSION "0.199.pre 7-Aug-2021"
+/*! \def MVERSION
+ * The current version of metamath.  It is incremented each time the software
+ * is modified.  When main versions are released, the version consists of a
+ * main version, followed by a dot and a three-digit subversion.  Intermediate
+ * versions are further followed by a free style suffix that should allow
+ * ordering.
+ * The version string is extracted and then processed by shell and perl
+ * scripts.  To avoid problems during replacements, refrain from
```
digama0, line 68:
```suggestion
 * scripts.  To avoid problems during replacements:
```

wlammen:
`Is it legal for Makefile.am.annotated to actually replace Makefile.am? `
Yes, it is.  The same holds for configure.ac.

wlammen:
`The build command is starting to look like a lot more than autoconf && ./configure && make. Shouldn't this stuff be part of the makefile?`
The build.sh prepares the input files configure.ac and Doxygen.diff.  This is a necessary step, before you can start the build process -- at least as long as you keep the version within metamath.c and not separately in an extra file. This cannot be part of make (is invoked too late), nor autoreconf (might be possible, but needs tons of patching and M4 stuff then.  I am not ready for that.)
`Also I think you aren't supposed to run ./configure every time.`  That's right, a simple call to the generated config.status yields the same results.  But then you cannot clean the build directory any more, and it is not a build-from-scratch.  The latter is the primary goal of the build.sh. Feel free to add another flag to the call that keeps essential build files.
I am aware that experts in this field can do a lot of optimization. And can write better shell scripts.  It is not my intention to provide a perfect, or even good solution.  But to give a framework, that we, or at least I, can write documentation in.  That is simple to install.
 Also I think you aren't supposed to run ./configure every time.

wlammen:
That's it for today

PR #52, opened 2022-01-30
Update build.sh

wlammen:
The build.sh script has been polished.  It does not support documentation build yet, but creates the executable from scratch in metamath-exe/build (by default) alone:
- supports input parameters;
- free choice of source and destination directories;
- extracts the version from C sources and displays it on request.  Also creates a metamath_version file during build, so that subsequent and external scripts have easy access to that value.
- Simplify script code by omitting the $BUILDDIR/ prefix on files.  This became unnecessary after issuing a cd $BUILDDIR.
- The grep regex for the MVERSION line was slightly flawed, and could potentially fail under very weird circumstances.

The unstable directory contains files still under development and changing frequently. Always use at your own risk.

Note the comment: `clear the build directory`. Files from previous builds are not honored (yet) (like make distclean)


wlammen:
I think one should break it down into 3 individual scripts: 
1. fetching the version from the C sources and printing it;
2. patching the version in configure.ac and Doxyfile
3. building (not installing) the artefacts in a build directory using 2.

jkingdon:
I'm a bit confused by putting this in "unstable". Is this script currently unused at all? Or are some features of it more unstable than others?

I mean, I've always just used the `cd src && gcc m*.c -o metamath` method from README.md but I'm not really sure what other people have been doing.

jkingdon:
> I'm a bit confused by putting this in "unstable". Is this script currently unused at all? Or are some features of it more unstable than others?

If the point is to move it out of `build` and reserve `build` for generated files, I'm all in favor of that. But maybe putting it in `src` or some place else is a better choice than `unstable`?

digama0:
I think the main build script should be in the root, and it can be a front end to anything else. If people see a `build.sh` file alone in the root they won't need to consult the documentation to guess that `./build.sh` builds the project, so we should try to make that true.

jkingdon:
> I think the main build script should be in the root, and it can be a front end to anything else. If people see a `build.sh` file alone in the root they won't need to consult the documentation to guess that `./build.sh` builds the project, so we should try to make that true.

I don't have super-rigid preferences but the above sounds good to me.

wlammen:
I put it into the unstable folder, since it might still not be fool-proof, and subject to major changes.  In particular not all of https://softwareengineering.stackexchange.com/questions/288205/what-are-the-advantages-of-build-scripts is covered yet.  IMO, it belongs to the src/ folder, but I was unable find anything else than "build/compile/install sequence" which supports this idea to some extent.

wlammen:
Obsoleted by #54

PR #53, opened 2022-01-30
Doxyfile

wlammen:
A Doxyfile template useable with doxygen.

Doxyfiles are possibly personalized and hence are never overwritten directly.  Instead a template file is added, that can be
- used instead of any previous Doxyfile by renaming it to Doxyfile;
- can be merged with an existing Doxyfile by prepending its contents to the existing one.  Since later settings overwrite previous ones when parsed, personal settings prevail that way;
- or merging both contents manually, laborious, but always suiting your needs best.

The suggested template is based on doxygen 1.8.17.  Default values are left out.  You can view the defaults by running `doxygen -g`.


digama0:
I'm wondering whether we should remove most of the comments from the Doxyfile template. The comments that are there look most useful for someone first setting up the template, but we should instead have comments that are relevant to someone that wants to change the docgen, so we should instead guide people to things they might want to change.

src/Doxyfile.diff
```diff
@@ -0,0 +1,84 @@
...
+
+#---------------------------------------------------------------------------
+# Configuration options related to the LaTeX output
+#---------------------------------------------------------------------------
+
+# If the GENERATE_LATEX tag is set to YES, doxygen will generate LaTeX output.
+# The default value is: YES.
+
+GENERATE_LATEX         = NO
+
```
digama0, line 84:
```suggestion
# diff to Doxyfile 1.8.17 with only default settings.

PROJECT_NAME = "Metamath Executable"
# TODO: Set to Metamath version
PROJECT_NUMBER = "Metamath-version"
PROJECT_BRIEF = "executable used to maintain *.mm files" 
PROJECT_LOGO = "Metamath.png"
OUTPUT_DIRECTORY = "../doxy"
# disable Graphviz .dot output
HAVE_DOT = NO
# disable LaTeX output
GENERATE_LATEX = NO
```

wlammen: The people likely editing this file are developers (like you) who want to tweak source documentation.  Say you like graphs in your output and have dot installed.  Why should this information `doxygen will assume the dot tool is
available from the path` be hidden from you?

Some of the changes I made in the past few days target the inexperienced developer, not the expert.  Annotating configure.ac for example.  Experts will likely be annoyed by the comments, that most often tell them nothing new and just clobber up the editor contents.  Yes.  I know.  But it is a matter of a few seconds to delete the comments and get a concise view.  By contrast, it is a matter of hours to wade through documentation hundreds of pages thick to find out what is going on for a newbie.  Statistics should reflect the probability of what kind your average reader is.  What do you think is the trade off then?

That said, I am not really picky about this point.  I can live with both results.

digama0: I was assuming that the dot setting was just a "no, this is inapplicable for us", and we don't need to duplicate doxygen documentation for someone who really wants to try an unsupported configuration. If you actually want to consider setting that to true, then sure, it would make sense to explain what else has to be done in order to turn that feature on.

I have nothing against documentation for the inexperienced developer, but it should be *our* documentation, recontextualizing the information in the context of this project. It's pretty obvious what `PROJECT_LOGO` does from the name and the value it is set to, and I think that it is actually much clearer to remove the wall of text spreading the three lines of interesting configuration across two pages.

wlammen: Apart from project name and number nothing is fixed so far in the doxygen configuration.  If others think my choices should be modified, yep, go ahead.  Note that the current state of affairs is **setting up** a framework useful to developers only (not end users).  While this set up shows points of commonality with the general build process and should be finally an integral part of it, I am not primarily bothered with the general build (apart from documenting it).  I want to get the doxygen documentation of the C sources off the ground.  To this end build.sh/Doxyfile.diff are convenience installations providing fast results in the current environment.  For testing, suggesting improvements and so on.  If they vanish after a few weeks and are replaced with something better - that is fine with me.

digama0: Okay. Frankly I don't know why you are spending so much time on doc setup and meta-documentation instead of actually writing documentation at this point. You don't need doxygen to get use out of doxygen style documentation - lean has something that resembles doxygen documentation but I have never seen it built or attempted to build it myself. It's still plenty readable as documentation.

I still think you should remove all the garbage from the doxygen.diff file. If you want to have comments in it, okay, but the comment ratio should not go beyond 2-1.

wlammen: >...why you are spending so much time on doc setup and meta-documentation instead of actually writing documentation at this point.

First, I guess, reviewing your code cost me about the same time.

Then, it is my sincere hope that I will not be the only one documenting the sources.  In private I had a system up for a few weeks now.  The problem is to get support of others, too.  You need to agree on a common style, find a way to work together, see where the limits are, and so on.  In the beginning I suggested AsciiDoc, now we use Doxygen.  This is a clear indication that my ideas need some review, and are not accepted out of the box.

The meta-documentation is just an example on how things can be done.  If people don't like the style, I hopefully learn soon.

As for the Doxyfile, I already accepted all of your suggestions.  I prefer getting on over disputing the comment style for more than a couple of hours.

wlammen: I thought resolving the conversation automatically accepts your suggestion.  It does not.  I manually updated the file now.

PR #54, opened 2022-01-30
Support doxyfile

wlammen:
Create a Doxygen documentation using build.sh.  For testing purposes.  TODO implement a make target doing the same.

This is - more or less - the state of pull request #49


jkingdon:
With respect to the build.sh changes I guess some of my bigger questions on that are over at #52 .

And I'm not sure I fully understood the `Doxyfile.diff` part - is the idea that there is a default configuration for "generate the documentation" (and, I suppose, a way for people to override with personal preferences) or is generating documentation supposed to be more manual? I will confess I probably didn't read through the build scripts enough to fully understand the mechanisms here.

The changes to src/metamath.c looked good to me - nice to see what we are actually hoping to do with all this stuff.

wlammen:
The Doxyfile may contain your personal settings.  For example you might want to have a Latex documentation rather than HTML files.  In such a case it would be annoying, if it gets overridden each time you update the source folder.  Another point is that doxygen itself is regularly updated, and the set of supported properties changes over time.  The .diff suffix suggests that it contains only differences to your current version whatever that is, that needs merging.  For starters, the .diff file can simply be copied as Doxyfile.

I am not committed to this particular handling.  If there are better ideas, go on.  I just don't want this block the whole process for longer.  It can be improved on iteratively.

digama0:
I think for the first cut we should not support custom configuration by users, since we're probably going to be serving the generated documentation ourselves anyway. If users want to add a diff on top of the existing file, that's their problem. Certainly we should not assume that *every* user has a Doxyfile; I would expect it to be the other way around, a minority of users have customization that they care enough about to expect to be able to apply to projects they find on the internet.

In short: does `./build.sh` with zero modification to existing files in the repo build the doc correctly? If so then it's fine.

jkingdon:
Although I don't think a lot of people will want to configure doxygen, I don't object to hooks which let them do so. I do think there should be a default configuration (in other words, I agree with "does ./build.sh with zero modification to existing files in the repo build the doc correctly?" as a goal).

wlammen:
I am inclined to make this configurable.  If you provide a Doxyfile per input parameter to build.sh, it uses the described merge mechanism, otherwise it relies on the Doxyfile as distributed,  Since your Doxyfile may be kept elsewhere, the Doxyfile name is not blocked any longer, and can be part of the source folder.

wlammen:
I had to copy the build.sh script from unstable to build to keep the CI happy.  This is not really intended, but I don't know how to solve the problem of the missing build.sh otherwise.

src/mminou.h
```diff
@@ -2,7 +2,12 @@
 /*        Copyright (C) 2020  NORMAN MEGILL  nm at alum.mit.edu              */
 /*            License terms:  GNU General Public License                     */
 /*****************************************************************************/
-/*34567890123456 (79-character line to adjust editor window) 2345678901234567*/
+/*34, encoded as a long;567890123456 (79-character line to adjust editor window) 2345678901234567*/
```
digama0, line 5: stray paste?

digama0 (commented):
I don't understand the semantics of the `-m` flag, nor why `build.sh` has to exist in both `build/` and `unstable/`. Suggestion: The `build.sh` file lives in the root, along with any other required build inputs like `Makefile.am` (although if we can stuff all this in `build.sh` then that works too). It is intended to be run from the root directory, and generates all files in the `build/` directory, by `cd`ing into the directory if necessary. No files should be placed in `src/` that are not source files, and the build script should not need to be run from that directory either.

The command line options look good, although I would prefer a more `make`-like input, e.g. `./build.sh doc`, `./build.sh clear`. The `help` and `version` commands can stay as `-h` and `-v` (although `--help` and `--version` should work too, if possible).

digama0:
Running `build/build.sh` from the build directory produced the following output for me:
```
build/> ./build.sh
cp: cannot stat 'Doxyfile.diff.orig': No such file or directory
rm: cannot remove 'Doxyfile.diff.orig': No such file or directory
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
<snip>
config.status: executing depfiles commands
make  all-am
make[1]: Entering directory '/home/mario/Documents/metamath/metamath-exe/build'
gcc -DHAVE_CONFIG_H -I.    -O3 -funroll-loops -finline-functions -fomit-frame-pointer -DINLINE=inline -g  -Wall -Wextra -MT metamath.o -MD -MP -MF .deps/metamath.Tpo -c -o metamath.o metamath.c
mv -f .deps/metamath.Tpo .deps/metamath.Po
<snip>
make[1]: Leaving directory '/home/mario/Documents/metamath/metamath-exe/build'
./build.sh: line 126: doxygen: command not found
doxygen not found. Cannot build documentation.
./build.sh: line 137: [: missing `]'
./build.sh: line 145: doxygen: command not found
```
* The `cp` and `rm` commands should be guarded by a test to ensure the file exists first, and should do something sensible otherwise.
* You should check for the existence of `doxygen` before attempting to call it
* the `[` error on 137 looks like a bash syntax error

wlammen:
1. If you empty the build folder, the Continuous Integration checks in Github let the PR not pass, see the failing commits like [provide a build.sh for CI](https://github.com/metamath/metamath-exe/pull/54/commits/d2df44c9bbea42f18b1e40d62feeeb6404fe9c19)
Apart from this keeping the build.sh duplicated is pointless, and it should be removed once the CI process is updated.
2. The semantics of -m (provide a path to the top directory) is to allow for running the build.sh script anywhere you like, instead of just in the build folder.
3. Your suggestions regarding the command-line parameters are welcome!  It is my intention to get this uploaded to find out how developers want to use it.
4. The existence of doxygen is checked in line 126.
5. Line 137 is flawed, you are right.
6. I did test the build.sh. I am going to do it now again. Just a moment...

digama0:
1. You can update the https://github.com/metamath/metamath-exe/blob/master/.github/workflows/build.yml file in this PR. The script is trivial: `./build.sh` from the `build` directory. My recommendation is to just remove the `working-directory` line so that it runs from the root directory.
2. That's fine but the default should be the same directory as the `build.sh` file itself. As I said. I want `./build.sh` from the root, with no arguments, to build metamath, exactly like the CI.
3. The PR is also for this.
4. The output of `command` has to be filtered or piped. See `run_test.sh` which does a similar thing for the `metamath` command.

src/mminou.h
wlammen: yes

wlammen:
1. Since the build.sh script moved to unstable it now clears the build folder completely (but nothing else, unless you use option -d unwisely)
2. cp: cannot stat 'Doxyfile.diff.orig': No such file or directory: source and destination were accidentally swapped
3. line 145: doxygen: command not found: If the previous test for doxygen failed, you should not call line 145

wlammen:
For now we have to live with unstable/build.sh (instead of root directory) until the build process has matured enough, and people know where build.sh should be (see my comment in #52)

wlammen:
> The output of command has to be filtered or piped.

Need to look at this closer.  But not tonight anymore.

digama0:
Could you move the build.sh file out of unstable? I know it's unstable, but it won't be when the PR is finished so you don't need to apologize for it in the PR itself. (Same thing for the "experimental" caveat in the script.)

digama0:
I will push some proposed changes to your branch, since github suggestions will not display them well. Feel free to modify or revert them if you disagree.

wlammen:
I moved build.sh to top dir right now, but didn't test it.

wlammen:
Test was successful now

digama0:
Okay, I pushed some commits. Summary of changes:

* Bash local variables are lowercase, and all variables have defined values at the start. (Otherwise, environment variables can change the behavior of the script, which you had an issue with before.)
* The `-d` option for the build dir has been renamed to `-o`
* The `-b` option skips autoconf
* The `-c` option clears the build dir; `-cb` will only clear, and `-c` without `-b` will clear and then recompile
* The new `-d` option enables doxygen. It is not on by default.
* The script no longer attempts to return to the initial directory. This is not necessary for shell scripts which are run like `./build.sh` because they run in a sub-shell.
* The autoconf patching is simplified to just `sed "s/REPLACED_BY_BUILD_SH/$version/g"`.
* The `configure.ac` and `Makefile.am` files are moved to the root, and `metamath.1` is moved to the `man` directory. (I tried clicking around popular repos at https://github.com/topics/c to see if there are any conventions here, and for autoconf projects both of these files tend to be at the root.)
* The `Doxyfile.diff` patching is done by appending instead of rewriting.

I didn't change the `Doxyfile` extension, but it seems to be more common to have a `Doxyfile` in the root rather than `Doxyfile.diff`. Indeed it seems backward that the user's doxyfile extension is called `Doxyfile` but the one provided by the project is `Doxyfile.diff`; I think the other way around would make more sense. The documentation in the doxyfile should also mention this behavior of the build script.

digama0:
I found a site that explains some of the project structure stuff around autotools projects: http://www.fifi.org/doc/autobook/html/autobook_46.html . It looks like I was wrong to put the `./configure` script in the build directory, because it is designed to be packaged in the source tarball such that `./configure && make` at the root works (or, if output files are being generated to a `build/` directory, `../configure && make`). The suggestion of using a `config/` directory to shuffle away the other files generated by autoconf is a good one; I'm going to try to set that up.

digama0:
Okay, I think I'm satisfied with the state of the build script now. Summary:

* `autoconf -i; ./configure; make` works, as does `autoconf -i; cd build; ../configure; make`
* The version number is extracted from the sources directly in autoconf (by calling `./build.sh`), so it works even if you bypass the script
* `./build.sh` works on a pristine installation as well as one with a previous compile. It moves the executable to the root of the project to make it easy to find.
* `./build.sh -d` runs the doxygen build, and has been tested to work. It produces html in the `build/html` directory

Autoconf really cares about things being in certain locations, so the file structure looks different than I had originally intended, but I'm happy to follow the lead of an opinionated build tool.

```
configure.ac
Doxyfile.diff
Makefile.am (stub pointing to src/Makefile.am)
build.sh
metamath (build output)
src/
  Makefile.am (source listing)
  metamath.1 (man page)
  *.{c, h}
doc/
  Metamath.png
  *.md
config/ (autoconf build output)
build/ (configure/make build output)
  html/ (doxygen output)
  src/
    *.o
    metamath <- removed from this location and put at the root
tests/ (test suite, unchanged)
```

jkingdon (approved):

digama0:
The bors merge bot has an invocation "bors d+" which means to delegate merging to the PR author. I'm doing that now - Wolf, hit the button when you're satisfied.

wlammen:
Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests.
I can only 'Close with comment'

digama0:
Oh oops, I thought you had write access. Would you like to?

wlammen:
Write access: Would you like to?

Today was the first time I actually needed it.  What obligations are tied to it?  If it helps getting around bottlenecks during holiday, its okay.  If I am expected to be in a permanent stand-by mode, possibly not.

I'd like to wait a couple of days for the fallout to settle, before touching this beast again.  But some early remarks:
1. Moving the built binary in its final location is usually part of the installation step.
2. If spaces in the version are a problem we should not use them. Actually IMO the version should be maintained in the configure.ac file and moved to config.h by autoconf, and finally transferred via an #include config.h  to the executable, but this is for later.
3. the build script can be split into independent parts like version.sh, gen-docs.sh and maybe even a do-build.sh where all the input checking and default handling is removed from.  This would be inline with having a separate run-tests.sh.  By the way there are distributions that do a self-test after build.  A separate scripts folder for stuff like that?

Where can we gather suggestions like this? On google groups?

digama0:
> Today was the first time I actually needed it. What obligations are tied to it? If it helps getting around bottlenecks during holiday, its okay. If I am expected to be in a permanent stand-by mode, possibly not.

Are you interested in being considered a "maintainer" for this project? That is, people might contact you with questions or bug reports about the code, usually via github issues. I don't think there are any minimum time requirements, but ideally there should always be someone to respond in a timely manner to reported issues and while I'm currently happy to fill that role it is always good to have someone else in case I'm not available for whatever reason.

> * Moving the built binary in its final location is usually part of the installation step.

I'm not sure I would call the root of the project "its final location". I would say that this build script doesn't have an "install" option, and if it did it would copy into somewhere more global. (Although, since the makefile is autoconf-built I believe you can do `make install` from the build directory and it will do something useful.)

> * If spaces in the version are a problem we should not use them. Actually IMO the version should be maintained in the configure.ac file and moved to config.h by autoconf, and finally transferred via an #include config.h to the executable, but this is for later.

Agreed that we should tackle this later. The specific issue that autoconf was having was that `make distcheck`, which produces a tar file with the name `metamath-VERSION.tar`, was using the name unquoted in some places. Based on the current look of the version name it seemed fine to just turn spaces to dashes to have something like `metamath-0.199.pre-2022-01-01.tar`, but we could also trim it down to just the version itself if needed.

> * the build script can be split into independent parts like version.sh, gen-docs.sh and maybe even a do-build.sh where all the input checking and default handling is removed from. This would be inline with having a separate run-tests.sh. By the way there are distributions that do a self-test after build. A separate scripts folder for stuff like that?

That all sounds fine to me, although certainly any other scripts should not go in the root. But if it simplifies things we can move stuff from `build.sh` into single purpose tools.

> Where can we gather suggestions like this? On google groups?

Right here seems fine. Or you can submit a PR doing some of this.

jkingdon:
> > Today was the first time I actually needed it. What obligations are tied to it? If it helps getting around bottlenecks during holiday, its okay. If I am expected to be in a permanent stand-by mode, possibly not.
> 
> Are you interested in being considered a "maintainer" for this project? That is, people might contact you with questions or bug reports about the code, usually via github issues. I don't think there are any minimum time requirements, but ideally there should always be someone to respond in a timely manner to reported issues and while I'm currently happy to fill that role it is always good to have someone else in case I'm not available for whatever reason.

We haven't been hugely structured on this (at least so far) but I think most regular contributors have write access at this point (although "regular contributor" is a larger group of people over at https://github.com/metamath/set.mm/ than here so I'm sort of including that in my answer). As far as I've noticed that hasn't come with any particular time commitment, just a general willingness to follow the norms in CONTRIBUTING.md and the like.

As for maintainer, we do have a list (currently just of @digama0 ) in the CONTRIBUTING.md file, and that's more as described above. There's still some ability to define one's own level of obligation/availability but maintainer is more obligation than write access.

(If anyone has a different vision of any of this, feel free to speak up. We've written down some of it, but some of the above are my impressions as much as anything "official").

Issue #40, opened 2022-01-10
Review


wlammen:
I reviewed recently some code, two pull requests among them were quite big.  We have neither code styles nor review check lists available.  Hopefully we decide on the documentation rules soon, so at least in theory we can fill these gaps.

Instead I followed the rules that are in effect in my company.  I think we should discuss the results here.  This all is about learning.  How can we improve the workflow?  What level of quality (both in code and review) should we aim at?  I am looking forward to hear what you have to say.

digama0:
I've been wondering how to approach this issue as well. Things will settle down eventually, once the code is a bit closer to locally optimal, but I expect to be putting in a large number of changes in the near future because the objective function has changed discontinuously and so we are currently far from that optimal point.

I don't want to be too forceful here in pushing my changes, but I think we should aim for higher throughput at the expense of review quality in the near term, at least until the big refactors are mostly done, set up CI and testing to ensure no obvious breakage, release a new version, and get some testing and bug reports from users in case of regressions.

Metamath has a lot of edge case behaviors, and I think we will need a big test suite to ensure compatibility with all of it (or at least the part of it we consider important). I think that good testing will be more useful in the long run than careful review (because it's a C project and eyeballs are only so good). Once we start to approach the optimum I think we can start to raise our documentation, testing, and review standards.

wlammen:
Just one answer in 10 days...

For comparison: Here is a code review check list: https://github.com/mgreiler/code-review-checklist.  Rest assured that I by far did not not use it exhaustively in my review attempts.  They were a compromise between my limited knowledge about the topic at hand, my limited technical skills, my limited time, the perceived eagerness of developers to make perhaps long-awaited changes, and general rules governing software development.

Of course, the overall goals of the metamath-exe project determines the amount of scrutiny used in review.  Like security, review rises costs, but may not yield any short-term advantages in return.  Review is about quality.  The advantages are often seen mid-term or long-term only.

IMO people voiced their expectation to have a rock-solid, reliable tool on hand.  This puts the level of quality very high.  My personal idea is to have a piece of software that is accessible to a wider audience.  This puts a limit on the complexity and diversity of the software as well.

My hope was to extract from the answers here a draft review check list.  So far this did not materialize.

digama0:
[...]

PR #55, opened 2022-01-31
add test suite and CI

digama0:
This implements the test runner described in https://github.com/metamath/metamath-exe/issues/40#issuecomment-1026006294, and adds two trivial tests.


tests/run_test.sh
```diff
@@ -0,0 +1,130 @@
+#!/bin/sh
```
wlammen, line 1: see https://askubuntu.com/questions/141928/what-is-the-difference-between-bin-sh-and-bin-bash

tests/run_test.sh
```diff
@@ -0,0 +1,130 @@
...
+
+usage() {
+  cat >&2 << HELP
```
wlammen, line 4: consider using quotes with the heredoc delimiter

tests/run_test.sh
```diff
@@ -0,0 +1,130 @@
...
+* '! expect_fail' means that we expect metamath to return exit code 1
+  instead of 0 on this input.
+HELP
+}
+
+if [ "$1" = "--help" ]; then usage; exit; fi
+if [ "$1" = "--bless" ]; then bless=1; shift; fi
+
+# Allow overriding the 'metamath' command using the METAMATH env variable
+cmd="${METAMATH:-metamath}"
```
wlammen, line 34: I consider this kind of parameter submission troublesome, because setting/unsetting this variable may differ both in location and time from its usage here.  Better submit it as a call parameter.

tests/run_test.sh
digama0: I know, I used `sh` on purpose. (I use `sh` over `bash` for scripts when possible because `sh` is more portable.)

tests/run_test.sh
digama0: I don't follow. This is the same pattern I used in all the other tools scripts.

tests/run_test.sh
digama0: I think it is conventional to use environment variables for command substitution, for example with the `$(CC)` variable in make to change the compiler. I will add both methods.

digama0: Actually I'm not sure I see any difference this way. There are two ways (now three) to use the script:

1. `./run_test.sh test`
2. `METAMATH=path/to/metamath ./run_test.sh test`
3. `./run_test.sh -c path/to/metamath test`

With (1), intended for maximum convenience, it will use `metamath` if you have it on your path, else you can set the METAMATH env variable (perhaps earlier in your session, perhaps in your `~/.profile` and long forgotten) to control where to find the program. This version has the downside that it can break due to misconfiguration: `./run_test.sh test` might mysteriously fail to work because METAMATH env var got set somehow. (Note that the script will give an error message in this case, so it shouldn't be too hard to diagnose.)

With (2), the environment variable is being set to something, so there is no possibility of misconfiguration. This is useful for scripts and CI, but it is more verbose. You always get the command specified.

With (3), you also always get the command specified. So it seems to have the same benefits and drawbacks as (2), except for a minor syntax change. So I don't really see the point of having both.

tests/run_test.sh
wlammen: I wonder whether the escape codes for the colours are portable...

tests/run_test.sh
wlammen: I meant cat << 'HELP'... This minor difference avoids the unnecessary scan for embedded $xxx variables.

tests/run_test.sh
wlammen: In our company we take care of a larger script project where the mentioned kind of "remote parameter injection" was routinely used.  This turned out to be a pain for maintenance, and was the cause for a lot of subtle bugs.

tests/run_test.sh
digama0: So the heredoc delimiter itself contains quotes? That seems confusing. I think I could also use a single quoted expression here, which would avoid the need for heredoc.

tests/run_test.sh
digama0: They are actually quite standard, but it's true that not every context interprets them. I considered adding a `--color=never/auto/always` like other binutils, but it's a bunch of extra code so I'll cross that bridge when someone complains about it.

tests/run_test.sh
jkingdon: `man bash` on my Ubuntu system says:

> If any part of word is quoted,  the  delimiter is  the  result of quote removal on word, and the lines in the here-document are not expanded.  If word is unquoted, all lines of the here-document are subjected to  parameter expansion, command substitution, and arithmetic expansion, the character sequence \<new‐line> is ignored, and \ must be used to quote the characters \, $, and `.

There is similar text in `man dash` so I suppose this is likely to be portable to other shells. It would appear to solve the problem you describe with backticks.

tests/run_test.sh
jkingdon: I vote for `/bin/sh` since other POSIX shells should work fine here. As for ANSI colors those aren't a property of the shell, they are a property of the terminal emulator or equivalent.

jkingdon (approved):

tests/run_test.sh
digama0: Oh, I didn't realize that was a special syntax. I still think it's weird, but if that's the standard thing then okay.

tests/run_test.sh
wlammen: This is a highly appreciated comment on https://stackoverflow.com/questions/484635/are-global-variables-bad.

"The problem with global variables is that since every function has access to these, it becomes increasingly hard to figure out which functions actually read and write these variables.

To understand how the application works, you pretty much have to take into account every function which modifies the global state. That can be done, but as the application grows it will get harder to the point of being virtually impossible (or at least a complete waste of time).

If you don't rely on global variables, you can pass state around between different functions as needed. That way you stand a much better chance of understanding what each function does, as you don't need to take the global state into account."

This, by the way, describes __exactly__ the maintenance problems of the script project in my company.

Or translated to the language of a software engineer:  You cannot assert the preconditions of the script any more, if you know nothing about the METAMATH environment variable, that may be accidentally set to something unexpected for whatever reason.  I think it is particularly difficult in a context, where you trust strict verification of mathematical proofs.  Convenience imo is NOT the highest priority, but predictability is.  

As for the shebang:  I don't think this is really a big issue.  I'd rather have this solved in a code style reading something along the lines:
"We allow using #!/bin/sh if the script passes a call with bash --posix.", accepted publicly by contributors X Y and Z.
In a review I can then simply look up whether the shebang matches this criterium, instead of unconditionally accepting a "I believe this won't hurt".

digama0: > Or translated to the language of a software engineer: You cannot assert the preconditions of the script any more, if you know nothing about the METAMATH environment variable, that may be accidentally set to something unexpected for whatever reason. I think it is particularly difficult in a context, where you trust strict verification of mathematical proofs. Convenience imo is NOT the highest priority, but predictability is.

As I said, if you want to get reproducible results then you take the global state out of the equation by using the `METAMATH=path/to/metamath ./run_test.sh test` or `./run_test.sh -c path/to/metamath test` forms, and you can see that this is what is used in CI. However, reproducibility is not the most important thing in all situations. Command line tools like `cargo` and `git` need to have good defaults and configuration at multiple levels in order to enable *convenient* use. Yes, this is global state, but it's there for a reason.

Regarding `/bin/sh`: this is already symlinked to `bash --posix` on my machine and this is the default on Ubuntu machines if not others. All testing of this script, including the CI, has been running it in this mode, so it should be easy for a reviewer to verify this.

wlammen: Not quite sure whether this is clearly understood: I don't deny a well-defined call is possible if certain restrictions are obeyed. My point is that the interface allows misuse, and, thus, is inherently unsafe.  This is either seen as a problem, or, well, not.

What should be clarified, maybe not in this thread, is whether I should continue reviewing at all.  The general task of a reviewer is described here https://en.wikipedia.org/wiki/Code_review.  I further published review check lists for discussion, and laid open that, without any rules established, (see #40 first post) I follow the guidelines of my company.  I am increasingly under the impression to deliver results others are not happy with.  And from my personal perspective things don't go well either.  My expectation is that any not openly flawed review comment should result in a code style, review check, or similar, that in future makes clear, how to proceed either in coding, testing or reviewing.  In the long run, we so could arrive at a well-tested development cycle, in which many persons can participate.  I don't see any progress here...

Just some thoughts popping off the top of my head.  I will post something to this respect in issue #40, I think, hopefully within the next week.

digama0: > My point is that the interface allows misuse, and, thus, is inherently unsafe.

The second half of this statement doesn't follow from the first half. There is a perfectly well defined contract for what's going on here. The environment variable is one of the inputs to that contract.

>  My expectation is that any not openly flawed and contentious review comment should result in a code style, review check, or similar, that in future makes clear, how to proceed either in coding, testing or reviewing. In the long run, we so could arrive at a well-tested development cycle, in which many persons can participate. I don't see any progress here...

I *did* make a change in response to this comment. I did not remove the environment variable because it is important to my workflow. Subsequently I have tried to argue for why it is there and we seem to have come to an impasse. I think we're not likely to make any more progress on this issue. If you don't trust the script, don't use it. You can verify that it is being used correctly in CI, and the other features are intended for well-meaning non-adversarial developers to be able to update the test suite.

Regarding the meta-point, I do believe our expectations are misaligned. I value your input, all of your reviews have been useful to me, but I get a general sense that you don't trust me or my work that negatively colors the review process. My experiences with [metamath-knife](https://github.com/david-a-wheeler/metamath-knife/) have gone more smoothly, even though it's mostly only @tirix who has the time to review.

wlammen: I think my last post was a lot motivated by my observation that I make close to no progress in the review and the documentation field.  IMO the review should not be governed by rules in effect in my company, that look quite arbitrary from a metamath standpoint.  But I am forced to.  Without any handout I can just follow my preferences.

These preferences in turn are based on my experience with code and where it badly fails in practice.  Relying on state not passed as a parameter is such a pitfall, and a contract given in form of a help text is a weak support.  I still don't understand why `run-tests -c "$METAMATH" <tests>` is not sufficient.

But again, my personal preferences should not be a critical factor in review.  So if a review rule states "The reviewer verifies that interfaces described by contracts hold.  S/he might annotate warnings if she detects signs of bad architecture, but nothing else.", then we wouldn't have this discussion.  Such a rule could be the spin off of this discussion, for example.

PR #56, opened 2022-02-06
refactor: getFullArg

digama0:
The changes here have been specifically itemized into small changes, most of which have a correctness justification. Reviewers are encouraged to look at the commits one at a time. This is in some sense an experiment in how to present refactoring changes in a way that isn't too overwhelming.

### DIFF
--- a/src/mmcmdl.c
+++ b/src/mmcmdl.c
@@ -1692,13 +1692,11 @@ static flag getFullArg(long arg, const char *cmdList1) {
     if (strcmp(tmpStr, tmpArg)) {
       printCommandError(errorLine, arg,
           "?A number was expected here.");
-      free_vstring(tmpArg);
-      free_vstring(tmpStr);
       ret = 0;
-      goto getFullArg_ret;
+    } else {
+      let(&keyword, str(val(tmpArg)));
     }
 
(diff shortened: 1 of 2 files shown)


src/mmcmdl.c
```diff
@@ -1770,6 +1775,7 @@ static flag getFullArg(long arg, const char *cmdList1) {
       if (!tmpFp) {
         let(&tmpStr,  cat("?Sorry, couldn't open the file \"", tmpStr, "\".", NULL));
         printCommandError(errorLine, arg, tmpStr);
+        free_vstring(tmpStr);
```
wlammen, line 1777: [move free of temps to just after last use]
Is there a good reason why free_vstring is applied to tmpStr here, but not to tmpArg, assigned in line 1742?  It is difficult for me to keep track of the stack.

wlammen:
It is a very good idea to split the commit into small chunks!

src/mmcmdl.c
digama0: No, you are right. This refactor has itself gone through some refactorings and that one seems to have been lost in the shuffle. It's supposed to be dropped on line 1746 just before it goes out of scope.

wlammen (approved):

PR #57, opened 2022-02-07
Add some of the tests from metamath-test

jkingdon:
- Does not include anything which took a long time to run
- Does not include versions of set.mm and the like - we check those
elsewhere
- Emphasis is on failure tests

Why copy them here rather than come up with some test rig to pull them from the https://github.com/david-a-wheeler/metamath-test repository? Well (1) as noted above not every test is included, (2) we very much want to be asserting on how the test fails - that a test produces a specific error message is a much better check that the code is still working than that it produces some kind of failure (perhaps unrelated), and (3) complexity in test scripts will .... well just isn't worth it here especially since metamath-test is fairly small and was last modified 6 years ago.

### DIFF
--- a/tests/anatomy-bad1.expected
+++ b/tests/anatomy-bad1.expected
@@ -1,0 +1,16 @@
+MM> READ "anatomy-bad1.mm"
+Reading source file "anatomy-bad1.mm"... 699 bytes
+699 bytes were read into the source buffer.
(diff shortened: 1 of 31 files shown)


jkingdon:
Oh CC @david-a-wheeler and thanks for the work in making metamath-test a thing!

digama0 (approved):

PR #58, opened 2022-02-08
Test disjoint

jkingdon:
As long as I've been saying "we should make a testsuite" I've been thinking in terms of making sure that all flavors of disjoint variable constraint violations are detected. It is one of the quickest way for a verifier to let invalid proofs slip through, or so has been my thought process (certainly when I'm writing proofs it is one of the most common places where the verifier catches me doing something which actually won't work).

Here's a start. My intended technique was to comment out the distinct variable checker (the lines in mmveri.c from `/***** Check for $d violations *****/` to `/***** (End of $d violation check) *****/`), writing a test, and then uncommenting out enough of the checker to get the test to now behave the same way as an unmodified metamath-exe. Well, I got as far as two passes (one for each of the two error messages - which turned into disjoint1.mm and disjoint3.mm in this pull request). No doubt it would be possible to poke at this more (either using this technique, or another one), but this is a start.

Along the way I made a somewhat surprising discovery which was that, without trying to do so, I found a test case which put the disjoint variable checker into an (apparently) infinite loop. I added this to the pull request as disjoint2.mm but the loopy bits (that is the portion which triggers the infinite loop) are commented out and so the .expected file just reflects the not-so-interesting parts of that file. The reason I say that it is the disjoint variable checker at fault is that I didn't see the infinite loop when I was working with a version of metamath with the disjoint variable checker commented out (as described in the previous paragraph).

### DIFF
--- a/tests/disjoint2.mm
+++ b/tests/disjoint2.mm
@@ -35,3 +35,9 @@ ${
   good $p |- ( ( x y ) ( z w ) ) $=
     xf yf combo zf wf combo ax-1 $.
 $}
+
+${
+  $d x z $. $d x w $. $d y z $. $d y w $.
+  stillgood $p |- ( ( x y ) ( z w ) ) $=
+    xf yf combo zf wf combo ax-1 $.
+$}
--- a/tests/disjoint2.mm
+++ b/tests/disjoint2.mm
@@ -1,0 +1,37 @@
+$c formula |- ( ) $.
+
+$v x y z w $.
+xf $f formula x $.
+yf $f formula y $.
+zf $f formula z $.
+wf $f formula w $.
+combo $a formula ( x y ) $.
+${
+  $d x y $.
+  ax-1 $a |- ( x y ) $.
+$}
+
+verybad $p |- ( ( x y ) ( z x ) ) $=
+  xf yf combo zf xf combo ax-1 $.
+
+${
+  $d x y z $.
+  stillbad $p |- ( ( x y ) ( z x ) ) $=
+    xf yf combo zf xf combo ax-1 $.
+$}
+
+semigood $p |- ( ( x y ) ( z w ) ) $=
+  xf yf combo zf wf combo ax-1 $.
+
+${
+  $d x y $.
+  $d z w $.
+  almostgood $p |- ( ( x y ) ( z w ) ) $=
+    xf yf combo zf wf combo ax-1 $.
+$}
+
+${
+  $d x y z w $.
+  good $p |- ( ( x y ) ( z w ) ) $=
+    xf yf combo zf wf combo ax-1 $.
+$}
--- a/tests/disjoint1.expected
+++ b/tests/disjoint1.expected
@@ -1,0 +1,19 @@
+MM> READ "disjoint1.mm"
+Reading source file "disjoint1.mm"... 140 bytes
+140 bytes were read into the source buffer.
+The source has 9 statements; 1 are $a and 1 are $p.
+No errors were found.  However, proofs were not checked.  Type VERIFY PROOF *
+if you want to check them.
+MM> Continuous scrolling is now in effect.
+MM> 0 10%  20%  30%  40%  50%  60%  70%  80%  90% 100%
+..................................................
+?Error on line 10 of file "disjoint1.mm" at statement 9, label "bad", type
+"$p":
+  xf xf ax-1 $.
+        ^^^^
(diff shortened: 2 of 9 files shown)


tests/disjoint2.mm
```diff
@@ -0,0 +1,20 @@
...
+combo $a formula ( x y ) $.
+${
+  $d x y $.
+  ax-1 $a |- ( x y ) $.
+$}
+
+$( Uncomment either good or bad to trigger an infinite loop
+   in the distinct variable checker. $)
+$( good $p |- ( ( x y ) ( z w ) ) $=
+  xf yf combo zf wf combo ax-1 $. $)
```
digama0, line 17: Is the name `good` indicative? Because I would have assumed that this is a DV failure - none of `x,y,z,w` are disjoint.

digama0:
Ah, this might not actually be an infinite loop. I think it's just waiting on input - the test harness should make sure to `SET SCROLL CONTINUOUS`

digama0:
The scrolling issue is fixed in #59 . This PR should probably get merged after that one

jkingdon:
> The scrolling issue is fixed in #59 . This PR should probably get merged after that one

Hmm I'll take a closer look a bit later. So this would happen when the output exceeds a certain size?

tests/disjoint2.mm
jkingdon: I realized that after I wrote it. You are correct that to be truly good this would need some $d added.

jkingdon:
> Hmm I'll take a closer look a bit later. So this would happen when the output exceeds a certain size?

OK, on a closer look this all makes sense. What I hadn't thought much about with respect to disjoint2.mm as opposed to disjoint1.mm or disjoint3.mm is that it has more disjoint variable violations and thus larger output.

Although it is entirely possible that we'll eventually pick #59 as the solution I don't think we should rush into that fix. That's a kinda large departure from how metamath-exe has always done things. Partly for that reason I updated this pull request to put SET SCROLL CONTINUOUS into disjoint2.in (putting it in the test harness wouldn't work unless we filtered "Continuous scrolling is now in effect." out of the output too and I wasn't immediately getting excited about making the test harness that clever).

tests/disjoint2.mm
jkingdon: OK, I apparently couldn't resist doubling down on this naming convention so the test now has `verybad` , `stillbad` , `semigood` , `almostgood` , and `good` .

digama0:
Besides #59, there are two alternate solutions:

* Add a command line flag or environment variable to change the default scroll mode, which would only be enabled in the test harness
* The test harness would add `SET SCROLL CONTINUOUS` to the start of the `.in` file, and wouldn't bother filtering the output; this would bloat all the `.expected` files and might make reading them more annoying though

My preference is for the #59 approach; lots of tools use isatty to go into "script mode" (the most common one I see is `--color=auto` in tools like `grep`, where you get colors if you use it from the terminal but not if you pipe to a file). I believe "tools mode" automatically enables continuous scroll for exactly the same reason, so I think this is consistent with the rest of the program, but it's true this is a breaking change and maybe there are some workflows that are affected. But I think all the usual ways to use metamath will continue to work: calling `metamath` or `rlwrap metamath` from the terminal results in prompted mode, and tools mode scripts were and still are in continuous mode.

jkingdon:
We can try to figure out the scrolling issue at #59 and #62 . Once we figure out what we want to do about that, I'll update this pull request.

jkingdon:
@digama0 (or anyone else) All tests are enabled on this branch and ready for re-review.

tests/disjoint2.mm
```diff
@@ -0,0 +1,37 @@
...
+  $d z w $.
+  almostgood $p |- ( ( x y ) ( z w ) ) $=
+    xf yf combo zf wf combo ax-1 $.
+$}
+
+${
+  $d x y z w $.
+  good $p |- ( ( x y ) ( z w ) ) $=
+    xf yf combo zf wf combo ax-1 $.
+$}
```
digama0, line 37:
```suggestion
${
  $d x y z w $.
  good $p |- ( ( x y ) ( z w ) ) $=
    xf yf combo zf wf combo ax-1 $.
$}

${
  $d x z $. $d x w $. $d y z $. $d y w $.
  stillgood $p |- ( ( x y ) ( z w ) ) $=
    xf yf combo zf wf combo ax-1 $.
$}
```

PR #60, opened 2022-02-08
refactor: parseCommandLine

digama0:
Continuation of the mmcmdl.c refactor.

### DIFF
--- a/src/mmcmdl.c
+++ b/src/mmcmdl.c
@@ -2004,7 +2004,7 @@ void parseCommandLine(vstring line) {
       }
       /* If character is comment, we're done */
       if (p == 0 && instr(1,tokenComment,chr(line[p]))) {
-        break;
+        goto parseCommandLine_ret;
       }
       /* If character is a special token, get it but don't change mode */
       if (instr(1, specialOneCharTokens, chr(line[p]))) {
@@ -2117,6 +2117,7 @@ void parseCommandLine(vstring line) {
     g_rawArgs++;
   }
 
+parseCommandLine_ret:
   /* Add length of command line prompt to each argument, to
      align the error message pointer */
   for (long i = 0; i < g_rawArgs; i++) {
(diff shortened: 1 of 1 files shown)


src/mmcmdl.c
```diff
@@ -2004,7 +2004,7 @@ void parseCommandLine(vstring line) {
       }
       /* If character is comment, we're done */
       if (p == 0 && instr(1,tokenComment,chr(line[p]))) {
-        break;
+        goto parseCommandLine_ret;
```
wlammen: continue (line 2003) and break are usual choices in a loop, goto is of ill repute

wlammen (changes requested):

src/mmcmdl.c
digama0: The `continue` is fine, but the `break` can't stay once the `switch` is introduced because it will target the switch instead of the loop. I would use labeled break if C had that, but it doesn't and `goto` is the next best thing.

src/mmcmdl.c
```diff
@@ -1989,13 +1989,18 @@ void parseCommandLine(vstring line) {
   /* mode: 0 means look for start of token, 1 means look for end of
```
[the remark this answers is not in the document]
digama0: It is removed in a later commit. The `#define`s themselves aren't even kept, it's just one step in the enum transformation.

digama0:
FYI: some of the transformations may look unmotivated if you read them in order. What actually happened is I did all of these transformations at once, and then broke it down into individual correctness-preserving transformations which you see in the commits. So some early commits are really just making it possible to do later ones without doing two things at once.

wlammen (approved):

wlammen (approved):

wlammen:
No, the presentation is fine.  Review is easy, but a second pass seems necessary to remove outdated review comments.

PR #61, opened 2022-02-08
bugfix doxygen not scanning source files

wlammen:
enrich documentation with information about *.c and *.h files.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -133,7 +133,7 @@ then
     exit 1
   fi
 
-  cd "$build_dir/src"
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -133,7 +133,7 @@ then
     exit 1
   fi
 
-  cd "$build_dir/src"
+  cd "$build_dir"
```
digama0, line 136:
```suggestion
```
We are already in the build_dir at this point

PR #62, opened 2022-02-09
Set continuous scroll in tests

jkingdon:
That is, have the test harness prepend "set scroll continuous" to
the .in file.

Update all .expected files to include "Continuous scrolling is now in effect."

This follows up on discussion in #58 and #59


digama0 (approved):
Still not sure this is the direction we should take, but the contents of the PR itself look fine.

jkingdon:
> Still not sure this is the direction we should take, but the contents of the PR itself look fine.

:shrug: Yeah, it isn't really a slam dunk for me either. I sort of go back and forth about whether my being wary of `isatty` is just my own aesthetic distate, or whether avoiding `isatty` for a little while until we make our mind is the thing to do, or whether it'll be fine and we'll come up with a Windows fix without too much trouble or whether we should do the Windows fix first (dropping Windows support, which I guess #59 does as currently written, seems like a big step) or what.

I guess we can give it a day or a few to see if more people weigh in - it has been less than 24 hours since we opened this discussion.

jkingdon:
OK, I'm glad to see some activity on #59 and #64 but the results there are..... well maybe I'll just say doesn't seem to me like we should rush into the #59 fix. Therefore, I will merge this at least for now. We can always revert later if we want.

PR #64, opened 2022-02-09
Add windows CI

digama0:
The tests fail with a segfault and `BUG #1502` though, so I should probably procure a working windows system to test. I used tinycc on a windows VM without problems, so I'm not sure what's wrong with the CI version.


digama0:
Oh boy, this is going to take a while to fix. The assumption that `sizeof(long) == sizeof(void *)` is pretty deeply ingrained in the code for `pntrString` and `nmbrString`.

wlammen:
Yep I know. In the upcoming doxy description of metamath this is described as a bug.  Affected is for example the structure nullPntrString.  A partial solution is using a union { long, void* }

digama0:
I'm working on a refactor that instead assumes `sizeof(void*) == sizeof(size_t)`. I am not so worried about uncommon architectures for which this assumption fails (in which case we would have to use `uintptr_t` instead), but win64 is not an out of date or out of support target.

PR #67, opened 2022-02-10
print output on failed tests

digama0:
This should help if, like the windows tests, something goes horribly wrong and the exit code is wrong, we now no longer suppress the output (except in the case of a run test, since in that case we never captured the output in the first place).

### DIFF
--- a/tests/run_test.sh
+++ b/tests/run_test.sh
@@ -86,11 +86,17 @@ for test in "$@"; do
     echo -n "test ${white}$test${off}.in: "
     result=$("$cmd" < "$test.in")
   fi
-  # exit code stored in $?
+  exit_code=$?
 
-  if [ $? -ne $expect ]; then
+  if [ $exit_code -ne $expect ]; then
     # If the exit code is wrong, the test is a failure and --bless won't help
-    echo "${red}failed${off}"; exit_code=1; continue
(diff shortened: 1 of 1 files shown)


tests/run_test.sh
```diff
@@ -86,11 +86,17 @@ for test in "$@"; do
...
     result=$("$cmd" < "$test.in")
   fi
-  # exit code stored in $?
+  exit_code=$?
 
-  if [ $? -ne $expect ]; then
+  if [ $exit_code -ne $expect ]; then
     # If the exit code is wrong, the test is a failure and --bless won't help
-    echo "${red}failed${off}"; exit_code=1; continue
+    echo "${red}failed${off} (exit code = $exit_code)"; exit_code=1
```
wlammen, line 93: I don't know what the exit_code is set for.  The continue forces the next loop, or exits it, if it was the last test. In the first case it is overwritten in line 89.  In the second case it forces the result of this script to 1, indicating a failure.

A failing test in the middle has no impact on the result of the script, only if it is last??

digama0: oh oops, there is a name clash there. The `exit_code=1` is for the exit_code of the script itself, while the `$exit_code` reference is the exit code of the metamath invocation of the current test. Will fix.

wlammen (approved):

PR #68, opened 2022-02-10
some corrections

wlammen:


src/mmdata.h
```diff
@@ -276,29 +277,36 @@ struct nullPntrStruct {
...
     long actualSize;
     /*! 
      * memory for a single void* instance, set and fixed to the null pointer.
+     * A null usually marks the end of the array.
      */
     pntrString nullElement; };
 /*!
- * \var g_PntrNull. Global instance of a memory block structured like a
- * \a pntrString, but fixed in size and containing always exactly one null
+ * \var g_PntrNull. Global singleton instance of a memory block structured like
```
digama0, line 299: It's not a singleton, in the sense that pointer comparison does not reliably detect empty `pntrString`s. https://github.com/metamath/metamath-exe/blob/1639113fe3ec39ea0bc57946f4c409671837c520/src/mmdata.c#L2499

src/mmdata.h
```diff
@@ -276,29 +277,36 @@ struct nullPntrStruct {
...
+     * A null usually marks the end of the array.
      */
     pntrString nullElement; };
 /*!
  * \var g_PntrNull. Global instance of a memory block structured like a
  * \a pntrString, but fixed in size and containing always exactly one null
  * pointer element.
  * 
- * \attention mark as const
+ * This structure must not be modified.
```
digama0, line 303: Can we not mark it as const?

src/mmdata.h
```diff
@@ -276,29 +277,36 @@ struct nullPntrStruct {
...
-     * currently used size of the memory block containing the \a pntrString.
+     * currently used size of the memory block containing the \a pntrString,
+     * excluding any hidden administrative data.
      * Note: this is the number of bytes, not elements!  Fixed to the size of a
      * single pointer element.
      */
     long actualSize;
     /*! 
      * memory for a single void* instance, set and fixed to the null pointer.
+     * A null usually marks the end of the array.
```
digama0, line 295:
```suggestion
     * A null marks the end of the array.
```
No need to vacillate.

src/mmdata.h
wlammen: The address of the nullElement is assigned to non-const PntrString. This leads to compiler warnings like

/home/wolf/metamath/src/doxy/src/mmdata.h:311:43: note: in expansion of macro ‘NULL_PNTRSTRING’
  311 | #define pntrString_def(x) pntrString *x = NULL_PNTRSTRING
      |                                           ^~~~~~~~~~~~~~~
/home/wolf/metamath/src/doxy/src/mmwtex.c:5102:3: note: in expansion of macro ‘pntrString_def’
 5102 |   pntrString_def(pntrTmp);
      |   ^~~~~~~~~~~~~~

I did not want to decide whether this can be resolved, so for documentation purposes I chose the weaker formulation.

digama0: One option would be to make `g_PntrNull` const but have a cast the const away in `NULL_PNTRSTRING`. That has the "advantage" of making modification to `g_PntrNull` UB instead of just breaking everything (which makes it marginally more likely to get a compiler warning or get caught in ubsan or something).

digama0: My suggestion is not to remove the "TODO: mark as const" in this case if you haven't conclusively determined that it is impossible.

wlammen:
Reverted the paragraph to the wording accepted in a former PR.

PR #74, opened 2022-02-22
Doxy

wlammen:


wlammen:
Documentation only

jkingdon (approved):

src/mmvstr.h
```diff
@@ -332,14 +389,82 @@ temp_vstring mid(const char *sin, long p, long l);
...
- *   this value is 0 and the length of sin in bytes.  Any value outside of
- *   these bounds is corrected to the closer one of these bounds.
+ *   this value is 1 and the length of __sin__+ 1 in bytes.  Any value outside
+ *   of these bounds is corrected to the closer one of these bounds.
+ * \returns a pointer to new allocated \a temp_vstring referencing the
+ *   requested portion, that is also pushed onto the top of \a tempAllocStack
+ * \post
+ *   A pointer to the substring is pushed on \a tempAllocStack, even if it
+ *   empty.
+ * \bug a stack overflow of \a tempAllocStack is not handled correctly.
```
digama0, line 399: Could you add a few examples of this and the other extraction functions? The 1-based thing always trips me up and it's easier to extrapolate from an example rather than read the description. I'm thinking something like `left("hello", 2) = "he"`

wlammen: I added attention tags to hint to the 1-based semantics.

There is a long example in the beginning of the file.  I considered to turn it into a doxygen page, so that it is accessible from there, too.

For now I added attention tags containing the requested example.

Note: Feel free to correct or amend any documentation.  This is mostly done after work, setting aside an hour or so.  I will continue at that pace, but support is welcome!

digama0: You are doing a good job, I have no complaints. Since this seems to be a somewhat "continuous" PR, just let me know when you would like me to merge it. (It looks like you added examples to all of the functions *except* this one, so I'm unresolving this as a reminder.)

My free time for metamath-exe comes and goes; right now I don't have much time for contribution but I can do reviews, so I'm happy if you keep up your current pace or anything else. I will remark if I see anything false or misspelled.

wlammen: Since I do not change the C source (apart from very obvious moves to get space for comments), you may check in intermediate states at any time.  Occasionally this will leave a comment in an unfinished or even incorrect state - well, yes.  I learn while writing, and every now and then I have to revisit my previous commits to correct a misunderstanding/misinterpretation or even an outright fallacy.  This won't last for long, I hope, and perhaps stimulate others to contribute a correction.

wlammen: I don't see the missing example.  left(...) hasn't got one, but it takes no index as a parameter.

digama0: Some basic tests would be useful simply as explication of the function behavior. Like `left("hi", 1) = "h"`, `left("hi", 3) = "hi"`, `left("hi", -1) = ""`

wlammen: I understand the usefulness of these examples in every day work, but that has not highest priority for me at the moment.  My primary goal is to deliver an overview over the current state of affairs, which means formally describing interfaces, semantics, pre-, postconditions, invariants and the like.  In the best case one can decide on how to transform the software without compromising trusted functionality.  Look at cmdInput for example.  It is way too complex, and has too many corner cases built in.  The documentation provides hints on how to improve on this situation.

left is a primitive, but has headroom for improvement, too.  The warning about missing UTF-8 compliance hints towards this.  Once you decide to address this issue, the examples become outdated, and need replacement/rework anyway.

Still, I understand your point.  I won't object to anyone modifying the doc in this respect.  I just want to focus on what I think is most pressing.  Forgive me for pursuing such poor standard.

jkingdon: > Since I do not change the C source (apart from very obvious moves to get space for comments), you may check in intermediate states at any time.

Hmm, I'm not exactly sure how best to fit this into how github handles pull requests, but :shrug: I guess "now" is an example of "any time" so I've merged this and you can open a new pull request for whatever you are working on next.

digama0:
It's not something I want to fix retroactively at this point, but for your next PR, please give it a better name than "Doxy", perhaps "add documentation in mmvstr" or so

PR #76, opened 2022-03-07
Convert comments to doxygen

digama0:
This will conflict with #75 to some extent, but I think it will be helpful for @wlammen to work from a base in which most functions are already (poorly) doxygen'd. In particular, the new doxygen comments should include or incorporate all of the text from the original documentation; there were some cases like `temp_vstring` or the mmvstr.h header where sizable comments were ignored when the doxygen comment was added, resulting in two comments with different syntax and somewhat overlapping contents.

### DIFF
--- a/src/mmcmdl.h
+++ b/src/mmcmdl.h
@@ -8,7 +8,8 @@
 #define METAMATH_MMCMDL_H_
 
 /*!
- * \file mmcmdl.h - includes for accessing the command line interpreter
+ * \file mmcmdl.h
+ * \brief includes for accessing the command line interpreter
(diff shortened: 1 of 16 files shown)


wlammen:
I am fine with that.  For one I learn a few techniques new to me, on the other hand it makes the documentation more useful for a wider audience.

wlammen (approved):
Only documentation is affected, no code change

jkingdon:
> work from a base in which most functions are already (poorly) doxygen'd

Surely the verb form of that should be "doxygenated"? :grin:

PR #83, opened 2022-05-08
Add test for missing $f

jkingdon:
This is based on an observation by Glauco on the metamath mailing list
and is taken from the relevant lines of set.mm.

### DIFF
--- a/tests/local-var-bad.expected
+++ b/tests/local-var-bad.expected
@@ -1,0 +1,23 @@
+MM> READ "local-var-bad.mm"
+Reading source file "local-var-bad.mm"... 290 bytes
+290 bytes were read into the source buffer.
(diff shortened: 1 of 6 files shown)


savask:
@jkingdon Sorry for posting in an already closed PR - but maybe it's worth adding these tests to https://github.com/david-a-wheeler/metamath-test ? Many verifiers use it as a test suite.

jkingdon:
> @jkingdon Sorry for posting in an already closed PR - but maybe it's worth adding these tests to https://github.com/david-a-wheeler/metamath-test ? Many verifiers use it as a test suite.

I certainly don't object but I'm not sure I know enough about metamath-test to do it. Like is there a need to specify a version of mmverify.py that it would pass for or that sort of thing.

Issue #33, opened 2022-01-05
Documentation


wlammen:
We need to improve on documentation.  And how it is done.

Visit https://github.com/metamath/metamath-exe/pull/32 and explore a possible solution.  Discussion started there, but IMHO it should be continued here.  Let me know what you think.

wlammen:
Are there any objections to follow these rules for documentation?
https://github.github.com/gfm/#introduction

I would like to come to a decision soon.

digama0:
That's a markdown spec, not documentation rules, but if I understand you correctly then please go ahead. Even in the worst case, if we want to change documentation format later we can do so without too much work. Writing the content is the hard part.

wlammen:
I deliberately chose a markdown spec to speed up the decision.

wlammen:
Based on this discussion I added a pull request adding a doc folder with a draft of simple documentation rules.  Please review.

PR #86, opened 2022-05-22
Fatal error handling

wlammen:
This aims at moving a basic layer of error handling to a separate place.  If you detect an out of memory condition, or even more severe, a corrupt memory state, there is little chance to recover from that state reliably.  So the best you can hope for, is to still be able to print a diagnostic message and then die.  All resources necessary for this last print should be pre-allocated, so chances are, this succeeds even in a heavily corrupted environment.  If not -- well, the program results weren't of use anyway, its just the user who is left puzzled on what the hell is going on.

Unrecoverable conditions like out-of-memory should have their program termination handling kept completely separate to minimize their dependency on the rest of the program state.  Metamath unfortunately does memory allocations and other complex operations under these severe conditions, for the sole purpose of some user convenience like paging and formatting.  This increases the chance of random program behavior, and creates circular dependencies.  Let's break this vicious cycle.

An infrastructure layer conforming to the above mentioned specifications with simple interfaces is provided here, documented and equipped with regression tests.  The layer is also UTF-8 ready.  This includes full ASCII compatibility.  A simple integration into the Metamath program is provided.


src/mmfatl.c
```diff
@@ -92,7 +92,7 @@ struct FatalErrorBufferDescriptor const* getFatalErrorBufferDescriptor()
  *   requires also information about available memory.
  * \param aDescriptor [null] a pointer to a \ref FatalErrorBufferDescriptor
  *   under investigation.
- * \returns 1 if acceptable, 0 else.
+ * \returns 1 if acceptabe, 0 else.
```
benjub, line 95: typo

wlammen: Still someone watching...  Thanks

benjub: Unfortunately, I cannot (yet) verify the correctness of the code...

benjub:
FYI several lines in several files are above 80 characters. I don't know how close you want to stick to this convention (and whether in all files or only in .c and .h files).

src/mmdata.c
```diff
@@ -3067,16 +3066,12 @@ long **alloc2DMatrix(size_t xsize, size_t ysize)
   long **matrix;
   long i;
   matrix = malloc(xsize * sizeof(long *));
-  if (matrix == NULL) {
-    fprintf(stderr,"?FATAL ERROR 1376 Out of memory\n");
-    exit(1);
-  }
+  if (matrix == NULL)
+    exitOnFatalError(__FILE__, __LINE__, "?FATAL ERROR 1376 Out of memory\n");
```
benjub, line 3070: Are these (1376, and below, 1377) magic numbers ? Or do they correspond to something ?

wlammen: The numbers are markers identifying the particular reason for a bug or an error.  A pair (__FILE__, __LINE__) could serve more or less the same idea.  This is the usual way to do, and it s part of the interface in my software here.

src/mmfatl.h
```diff
@@ -0,0 +1,231 @@
...
+ * is constructed.  In our context, this buffer is pre-allocated, and fixed in
+ * size, truncation of overflowing text enforced.
+ *
+ * Documentation
+ * =============
+ *
+ * Most comments are written in <a href="https://doxygen.nl/index.html">
+ * doxygen</a> (Qt variant) style.  If you have doxygen installed on your
+ * computer, you may generate HTML documentation out of them with its root
+ * page placed in build/html/index.html by running build.sh with the -d option.
```
digama0, line 53: Why is this here? Is the general documentation style related to fatal errors in some way?

src/mmfatl.h
```diff
@@ -0,0 +1,231 @@
...
+/*--------------------   Allocation Of An Error Buffer   --------------------*/
+/*!
+ * Basic parameters controlling pre-allocation of basic routines and data.
+ * Pre-allocation helps to construct a diagnosis message, even in a
+ * corrupt environment.
+ *
+ * A pre-allocated buffer is provided to allow embedding data in a diagnosis
+ * message.
+ */
+struct FatalErrorBufferDescriptor {
```
digama0, line 116: IIRC I don't think metamath.exe uses uppercase for type names, I think they are snake case.

src/mmfatl.c
```diff
@@ -0,0 +1,1591 @@
...
+#   define FREE free
+#   define MALLOC malloc
+#endif
+
+/*!
+ * C has no boolean type, instead an int is used.  To ease code reading we
+ * encode one boolean value here.
+ * int cond = FALSE; if (cond) { } skips the code block.
+ */
+#define FALSE 0
```
digama0, line 34: What the heck is this? Don't make macros for true and false, use the appropriate imports and use `true` and `false`.

src/mmfatl.c
```diff
@@ -0,0 +1,1591 @@
...
+/*!
+ * C has no boolean type, instead an int is used.  To ease code reading we
+ * encode one boolean value here.
+ * int cond = TRUE; if (cond) { } branches into the code block.
+ */
+#define TRUE 1
+
+/*! terminating character of a text string never used elsewhere */
+#define NUL '\x00'
```
digama0, line 43: I don't think this is preferable to just using `0`. It's only used once anyway.

src/mmfatl.c
```diff
@@ -0,0 +1,1591 @@
...
+ * character of a completely transferred sequence.  ASCII encoding is a single byte
+ * subset encoding of UTF-8 that always yields the last transferred byte.
+ * \param begin [not null] a pointer to the buffer start.
+ * \param last [not null] a pointer to the last transferred byte in the buffer.
+ * \returns a pointer to the last valid byte, or NULL if it does not exist.
+ *   The latter can happen only if either your buffer is less than 4 bytes
+ *   long, or if it contains invalid UTF-8 encoding.
+ */
+
+static char* findLastUTF8Byte(char const* begin, char* end)
```
digama0, line 72: We don't support UTF8 in the first place, what is this?

src/mmfatl.c
```diff
@@ -0,0 +1,1591 @@
...
+ * \param aDescriptor [null] a pointer to a \ref FatalErrorBufferDescriptor
+ *   under investigation.
+ * \returns 1 if acceptable, 0 else.
+ */
+static int isValidFatalErrorBufferDescriptor(
+    struct FatalErrorBufferDescriptor const* aDescriptor)
+{
+    return aDescriptor != NULL
+        && aDescriptor->size > 0
+        && aDescriptor->ellipsis != NULL? TRUE : FALSE;
```
digama0, line 135: I can't tell from this indentation whether you want everything in the conditional or whether the last conjunct is supposed to be `aDescriptor->ellipsis != NULL? TRUE : FALSE`. Also `x ? TRUE : FALSE` is bonkers, that's just `x`. (Everything here is a proper boolean value so it can't be that you want to convert an int to a bool.)

src/mmfatl.c
```diff
@@ -0,0 +1,1591 @@
...
+    }
+    return  result;
+}
+
+char const* getFatalErrorMessage()
+{
+    return memBlock? fatalErrorBufferBegin() : NULL;
+}
+
+/*===================    Parsing the Format String   =====================*/
```
digama0, line 254: And why are we implementing `printf` by hand now?

digama0 (changes requested):
This is a crazy amount of code to be adding for something that should just be `printf(); abort(1);` It's also in an area that gets very little testing and is difficult to exercise even deliberately, so I would prefer something as short and straightforward as possible. Not to mention that this doesn't even address the most important purpose of fatal error handling, namely to prevent loss of data in the event that there are unsaved changes in memory.

We should be making the program smaller and cleaner, not adding entire new modules just to solve a problem that isn't a problem.

src/mmfatl.h
wlammen: To my knowledge no general documentation generation is installed throughout Metamath.  I added some support of doxygen a couple of months ago, but that is by no means finished.  This is contrary to the situation here: Both the interface here and its implementation in mmfatl.c is documented.  Once this holds for all other mm* files, AND a common style is agreed upon, I am happy to move this into a central location.

src/mmfatl.h
wlammen: Ok, this was my choice, but I do not insist on it.

src/mmfatl.c
wlammen: I don't know how to interpret https://en.cppreference.com/w/c/types wrt to bool/true/false.  For one they are C99, and then they look deprecated.  But I can certainly remove the macros here, if considered necessary.

src/mmfatl.c
wlammen: Future extension.  I'd like Unicode support all over Metamath in the end.  You need to start somewhere.

src/mmfatl.c
wlammen: Not exactly.  C discriminates between 0 and nonzero, but not between 0 and 1.

src/mmfatl.c
wlammen: Because I don't know what printf does, when you are in an out-of-memory situation, or even worse, your memory is corrupted.

src/mmfatl.c
digama0: They are hardly deprecated; to the contrary the reason they are going away in C23 is because the *macro* is going away to make way for [the keyword](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2655.pdf) with the same semantics. C has clearly taken steps toward making `bool` / `true` / `false` the preferred spelling and `stdint.h` is how you should use it in C11 (which we are using).

src/mmfatl.c
digama0: What does that mean? The relational operators [are defined to always return 0 or 1](https://stackoverflow.com/a/37292153/890016).

src/mmfatl.c
digama0: I don't think we should be reinventing aspects of the C standard library here. This is just more code to have to deal with for little gain. If this was actually fixing a present bug I would understand, but the harms here are entirely speculative.

src/mmfatl.h
digama0: Then you should do it in a separate PR and you should do it uniformly to everything.

wlammen:
> We should be making the program smaller and cleaner, not adding entire new modules just to solve a problem that isn't a problem.

The code here emulates some of what Metamath implements currently, which includes pagination and a lot of other user convenience, even when it is about to exit because of severe memory problems.  This is a flawed design.  I'd never come up with an implementation like this, if I felt not obliged to keep the functionality to some degree.  The module is designed to allow replacement because it communicates with the rest of the program through a small interface.  The complexity is under the hood, and it is reduced, maybe not to your favorite degree.

src/mmfatl.h
digama0: > To my knowledge no general documentation generation is installed throughout Metamath. I added some support of doxygen a couple of months ago, but that is by no means finished. This is contrary to the situation here: Both the interface here and its implementation in mmfatl.c is documented. Once this holds for all other mm* files, AND a common style is agreed upon, I am happy to move this into a central location.

We [already discussed this before](https://github.com/metamath/metamath-exe/issues/33), and we agreed to use Doxygen, so there isn't any reason to hedge about this. If you want to document that you are using doxygen, it should be in the README.md or CONTRIBUTING.md anyway. It's entirely unnecessary to document that the code is documented in the documentation -- we can see that.

> Further note that this header file exports 4 simple routines dedicated to pre-allocation and deallocation of memory needed for formatting under extreme conditions, one formatter (like printf) and one exit call (close to abort). That's all. The criticized complexity of the implementation is hidden behind this interface, and is not exposed. Feel free to replace it with a handful of printf and abort lines.

Can you send me a PR with this? That sounds a lot easier to merge, and it's no worse than what the code is already doing (indeed, it fixes the main identified actual bug, which is that `outOfMemory()` is re-entrant).

src/mmfatl.c
wlammen: I did not know that.

src/mmfatl.c
wlammen: Hmm.  The goal is to guarantee Metamath verifies proofs as close to 100% as possible.  This covers reliable error detection and handling.  I could answer 'Assuming no harms is entirely speculative' as well.

src/mmfatl.c
wlammen: I did a quick lookup at the time of writing, and decided to use my macros as a replacement.  Looking closer now I too think we should use the standard keywords.

src/mmfatl.c
wlammen: My habit

src/mmfatl.c
digama0: > Hmm. The goal is to guarantee Metamath verifies proofs as close to 100% as possible. This covers reliable error detection and handling.

While I agree in the abstract that we should be improving the reliability (and more importantly, maintainability) of metamath.exe, I'm not sure I agree with the extreme phrasing of "as close to 100% as possible". If we're talking metamath in general and not metamath.exe, then I believe the best way to get verification as close to 100% as possible (and here I do agree with the claim) is to not implement the verifier in C at all, and use some formal methods to construct the verifier and/or make use of really small verifiers that can easily be eyeballed for correctness.

But for metamath.exe, it is much more than a verifier, and it is also not a verifier on the critical path for the aforementioned reasons, so keeping the code under control is a much higher priority than correctness except in a few key places. That's not to say that correctness isn't important, but rather that it is one of multiple goals. It's not at the level where I would accept a bugfix for a theoretically predicted bug that makes the code more complicated and less clear.

> I could answer 'Assuming no harms is entirely speculative' as well.

We have a bug tracker if harms become more than speculative.

wlammen: > The use of format strings is entirely a convenience, and it's not so convenient if we have to maintain the parser too.

Exactly.  If you give up on the idea of formatted error messages in a tight or corrupt memory situation, you can dispose of half the code in this PR.  This is a sensible design decision.  But it breaks with the current implementation.

As for the parser: It has three states, so about 9 transitions (including operations not changing state) to watch.  This is really not out of the world.  The correctness is supervised by the regression tests.  I could dig out bison if this helps in such a small case.  My guess is: You have some pain checking the code in now, but I expect it to be stable for years afterwards.

And yes, radically stripping down complexity is an option, if not a necessity, judging by the resources the community is currently allocating.  Important questions.  Beyond my responsibility here.

src/mmfatl.h
wlammen: The hint towards doxygen should not be part of a doxygen comment.  This should be corrected.

src/mmfatl.c
```diff
@@ -0,0 +1,1591 @@
...
+    resetParserState(&state, messageFormat);
+
+    /* marks the current end position in the buffer */
+    state.buffer = setLocationData(file, line);
+
+    /* now process the message */
+    va_start(state.args, messageFormat);
+    appendMessage(&state);
+    va_end(state.args);
+    fprintf(stderr, getFatalErrorMessage());
```
wlammen, line 618: Check whether fprintf is usable here

src/mmfatl.h
wlammen: I've looked into mmdata.h, and it is more than 90% camel case and less than 10% snake case.  Here are examples:
processUndoStack, free_pntrString.  Is there a howto somewhere giving a style guide?

digama0: Those are functions, not types.

If the style is inconsistent, then I'm happy to change it, but like I said that should be done in a PR which is only about that and is implemented across the board, not just in a few files.

src/mmfatl.c
wlammen: I don't know what to do here further.  The goals of this PR are written down under the title 'Rationale' in mmfatl.h, their implementation here is a standard conformant one.  The goals can be extended or reduced, but that is not this PR any longer then.

src/mmfatl.h
wlammen: Whatever this takes it is now deferred to a later contribution.

wlammen:
I don't know what to do here further.  The goals of this PR are written down under the title 'Rationale' in mmfatl.h, their implementation here is a standard conformant one, and the complexity seems to be of the order required for them.  The goals of course can be extended or reduced, but that is not this PR any longer then.

digama0:
Can you add a test case which exercises the code? If you could break off the part that adds regression testing and fixes known bugs without adding significant new code into a separate PR, this will be a lot easier to review. As for this PR, I don't have the time to review it well enough to convince me that it is all working as designed, especially in the absence of tests.

wlammen:
If you run `./build.sh -ct && ./metamath` a regression test is executed on startup, and the result is summarized at the end.  I used it yesterday to verify the transition to bool did not introduce an error.  This regression test calls all the internal and external functions under various conditions and compares the results to expected outcome.

What I can do is to create a new PR where I introduce commit by commit each time a new function accompanied by its regression test, and in the end it adds up to this one.  Is this ok?

digama0:
This should be in the CI

digama0:
Adding functions one at a time will definitely make this PR easier to review

wlammen:
Separating the regression tests from the code and integrating it to CI can be the final step.  Ok then.  I need some time to do prepare this.

digama0:
I would like each PR to be merged along with a test pertinent to it, so adding all the tests at the end doesn't sound like a good idea.

wlammen:
Each commit will both add a function/primitive feature along with a test.  This test won't vanish afterwards, and will always be invoked in a test run.  Over time test code and functionality will grow.  The diffs will be nice and small.

wlammen:
This PR will be rebuilt more systematically in a different one.

PR #88, opened 2022-06-05
Add example of a non-unique grammar from the mailing list.

jkingdon:
So I'll confess that I've written metamath proofs for a long time without really digging into syntax theorems so I'm not really the person to comment on what this means.

But I saw that @digama0 posted this example to the mailing list and it looked like the kind of thing we want in our testsuite.

Feel free to suggest a better comment, or other ways this could be different.

### DIFF
--- a/tests/nuniq-gram.expected
+++ b/tests/nuniq-gram.expected
@@ -1,0 +1,10 @@
+MM> READ "nuniq-gram.mm"
+Reading source file "nuniq-gram.mm"... 449 bytes
+449 bytes were read into the source buffer.
+The source has 7 statements; 3 are $a and 0 are $p.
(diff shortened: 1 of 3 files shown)


digama0:
I don't mind having this in the test suite, but keep in mind that metamath.exe does not really interact with the counterexample-like aspects of this example at all, because it doesn't do grammatical parsing (except in MM-PA to some extent).

jkingdon:
> I don't mind having this in the test suite, but keep in mind that metamath.exe does not really interact with the counterexample-like aspects of this example at all, because it doesn't do grammatical parsing (except in MM-PA to some extent).

Hmm, OK. I suppose some day maybe we'll have this example or one like it in a test for MM-PA or mmj2 or (whatever), but I guess there's nothing wrong with this as a first step.

PR #89, opened 2022-06-07
Plug 2 vstring leaks

jamesjer:
Valgrind found both of these.  In the mmcmdl.c case, tmpStr can point to the allocated memory in the else branch of the if, as well as the then branch, so move free_vstring after the if.  In the mmpars.c case, there was no free_vstring invocation at all.


wlammen:
It's right.  Both corrections describe memory leaks.  The free_vstring deallocaton are inserted/moved right after their last use, so they are best guarded against exceptions that might occur within allocation and deallocation.
Even should exception safety not perfectly be met, I support this pull request, since it is an improvement over the existing code.
Congratulation to apply Valgrind to this code, @jamesjer.  I used it several years ago the last time in a different project.  Nice to have someone in the team that knows about it AND uses it.

wlammen:
Approved

digama0 (approved):

PR #92, opened 2022-06-27
Add hypothesis and dv usage qualifiers to VERIFY PROOF

jamesjer:
Regarding the discussion at https://github.com/metamath/set.mm/issues/2648, here is one way metamath-exe could be enhanced to show some unused elements of metamath databases.  This PR adds the following qualifiers to `VERIFY PROOF`:

- /DISJOINT_VARS checks for unused non-dummy dvs
- /ALL_DISJOINT_VARS check for all unused dvs
- /ESSENTIAL checks for unused $e statements
- /FLOATING checks for unused $f statements
- /HYPOTHESES is /ESSENTIAL and /FLOATING combined


wlammen:
Can look into this, but probably not before weekend.

src/mmcmds.c
```diff
@@ -4674,8 +4674,13 @@ void eraseSource(void)    /* ERASE command */
 } /* eraseSource */
 
 
-/* If verify = 0, parse the proofs only for gross error checking.
-   If verify = 1, do the full verification. */
+/* If verifyFlag = VERIFY_NONE, parse the proofs only for gross error checking.
+   If verifyFlag includes VERIFY_FULL, do the full verification.
```
wlammen, line 4678: It seems this flag only controls a progress bar and a couple of result print outs. Please be more specific.

wlammen (changes requested):

jamesjer:
I haven't finished doing everything you requested yet.  In particular, I have not done any refactoring of verifyProof.  I'm pushing an update anyway so you can see my progress.  I am traveling starting tomorrow, so won't get back to this probably until July 11 at the earliest.

Note that I removed checking for unused $f statements, since that doesn't seem to make any sense.

wlammen:
Hmm.  trujust is an example with local $f instructions.  Were something like vz.tru $f setvar z $. added, that would be a redundant floating hypothesis.  I thought you had such a scenario in mind.

jkingdon:
How does the distinct variable part compare to https://github.com/metamath/set.mm/blob/develop/scripts/check-dvs.py ?

jamesjer:
I apologize for leaving this hanging.  @jkingdon the script in question checks for variables that aren't mentioned at all in the theorem or proof.  This PR, on the other hand, shows you dvs that the verifier never accessed while checking the proof, which implies that the verifier would still succeed if the dvs weren't present at all.

@wlammen , that would be useful, but the code in the original version of this PR checked for unused $f statements globally, which produced voluminous, and not at all useful, output.  I'll have to think about how to constrain it to local $f statements.

src/metamath.c
```diff
@@ -6313,11 +6313,15 @@ void command(int argc, char *argv[]) {
     }
 
     if (cmdMatches("VERIFY PROOF")) {
-      if (switchPos("SYNTAX_ONLY")) {
-        verifyProofs(g_fullArg[2],0); // Parse only
-      } else {
-        verifyProofs(g_fullArg[2],1); // Parse and verify
-      }
+      i = (switchPos("SYNTAX_ONLY") != 0) ? 0 : 1;
```
benjub, line 6316: Here and below, isn't our convention to write conditions as `(switchPos("SYNTAX_ONLY"))` instead of `(switchPos("SYNTAX_ONLY") != 0)` ?

GinoGiotto:
I heavily used these functionalities to optimize https://groups.google.com/g/metamath/c/OB2_9sYgDfA and they are quite amazing. While the [Python script](https://github.com/metamath/set.mm/blob/develop/scripts/check-dvs.py)  found [this PR](https://github.com/metamath/set.mm/pull/3699) as result (which is about 50 or so eliminable dv conditions), the /DISJOINT_VARS option currently finds 14,406 dv conditions that can be eliminated in set.mm.

Only problem is that the /FLOATING and the /HYPOTHESES options currently do not work @jamesjer and they don't even appear in the list of valid options.

But once everything is settled I'm strongly in favour of merging this PR.

src/metamath.c
jamesjer: Yes, you are right.  I have changed the code accordingly.

jamesjer:
@GinoGiotto Thank you for the comments.  I will note that we should be cautious about removing anything based on these checks.  Some "unused" dvs and essential hypotheses are there on purpose, even if the verifier doesn't use them when verifying.  For example, ax6v has an unused dv that is the entire point of ax6v, and a1ii has an unused $e on purpose.

GinoGiotto:
> Some "unused" dvs and essential hypotheses are there on purpose, even if the verifier doesn't use them when verifying.

I know :-)

Indeed I was thinking that an exception list might be a good idea. For dv conditions a good rule of thumb could be to exclude checking those theorems with a label `ax*v`, which are usually weakened to show that the original axiom `ax-*` was provable from them. A quick search shows that the follwing label list [ax6v](https://us.metamath.org/mpeuni/ax6v.html), [ax7v](https://us.metamath.org/mpeuni/ax7v.html), [ax7v1](https://us.metamath.org/mpeuni/ax7v1.html), [ax7v2](https://us.metamath.org/mpeuni/ax7v2.html), [ax8v](https://us.metamath.org/mpeuni/ax8v.html), [ax8v1](https://us.metamath.org/mpeuni/ax8v1.html), [ax8v2](https://us.metamath.org/mpeuni/ax8v2.html), [ax9v](https://us.metamath.org/mpeuni/ax9v.html), [ax9v1](https://us.metamath.org/mpeuni/ax9v1.html), [ax9v2](https://us.metamath.org/mpeuni/ax9v2.html), [ax12v](https://us.metamath.org/mpeuni/ax12v.html), [ax13v](https://us.metamath.org/mpeuni/ax13v.html), should be comprehensive enough for now.

icecream17:
Perhaps we could have a list of such theorems in the set.mm repo. This way, if someone wants to add extra DVs or hypotheses in a mathbox theorem for some reason, metamath.exe wouldn't have to change, and people wouldn't need to reinstall it.

I am very interested in this pr; both the DV and hypothesis checker seems very useful

jkingdon:
> Perhaps we could have a list of such theorems in the set.mm repo.

I think this would be required if we want to move forward with this (perhaps in $j directives or some similar mechanism).

I can think of a few examples in iset.mm which aren't in the list posted in this issue and I expect we'll add more in the future.

tirix:
I'm probably often using too many DV conditions, this seems kind of wrong, but since this is completely unchecked, there is no easy way to fix this, and this PR would help.

GinoGiotto:
> I think this would be required if we want to move forward with this (perhaps in $j directives or some similar mechanism).

Using $j directives is a good idea, they wouldn't be invasive since there probably aren't many theorems to be excluded. If there aren't other known issues, I encourage moving forward with this PR.

In set.mm there is also [axsep](https://us.metamath.org/mpeuni/axsep.html) to be added in the list.

benjub:
The $j-mechanism sounds good. So, there are two cases, like
* `$( $j intentional-hypothesis 'blabla'; $)`
* `$( $j intentional-DV-condition 'x y A'; $)`

?

For the second item, since DV conditions do not have labels, there is no canonical way to refer to them in a $j-statement. For instance, if I write `$d y x $.` (resp. `$d y A x $.`), will a mention to `'y x'` be accepted ? Or should it be exactly the string `'y x'` (resp. `'y A x'`) and nothing else ?

Also, in both cases, some hypotheses or DV conditions may be intentional for some theorems and not others, a situation which may arise in the case of long scopes. This is one more reason to avoid long scopes, which are error-prone.

icecream17:
I think `$j` DVs should be able to just list what intentional extra DVs there are, like normal DVs, independent of however the actual DVs are written

Issue #10, opened 2021-12-24
Speed up minimize command


savask:
About an year ago I sent Norman a proposal on how to improve the performance of the minimize command in metamath-exe, and by the suggestion of @digama0 I'm reproducing this letter here:

>Hello, Norman.
>
>I have an idea which can potentially speed up "minimize" command in metamath.
>
>Correct me if I'm wrong, but the general "minimize" command goes through all theorems/axioms in the database and tries matching them against the proof (by calling minimize_with). The idea is that one needs to consider only statements which only contain constants found in the proof of the minimized statement. For example, "mideulem" does not use the constant CC in its proof, so the whole bulk of theorems on complex numbers need not be checked.
>
>So I suggest to do as follows. Before minimizing the statement one should scan through its proof and collect all constants used (I think that already appears in "syntax hints"). Next, all statements should be checked against this collection, leaving those statements which use only constants from the collection. Finally, the forward-backward minimization loop can begin its work with the only change that now it should use only statements which passed this "validation" step.
>
>Clearly this can be done very effectively by storing collected constants and statements in set-like structures (or at least in sorted lists).
>
>I made a simple tool which shows constants used in the proof and the number of statements using only these constants. For example, mideulem uses only the following constants (it's a Haskell list):
>
>["class","TarskiG","=","setvar","pInvG","->","`","|-","Itv","=/=","dist","<->","wff","leG","Base","e.","(","/\\","E.","LineG",")","perpG"]
>
>And that results in only 2080 out of 37912 statements being relevant for minimization (that's ~5.5%). For "fourierdlem103" there are only 8821 relevant theorems (~23%). Haskell is arguably not the fastest language, yet this pre-minimization step went in under a second.
>
>To be honest, I spent some time studying metamath.exe code and while it is pleasant to read (thanks to the abundance of comments), it is certainly not a trivial task to modify it :-) I don't think I will be able to make such a change myself, so I decided to report it to you instead; I hope you will find it interesting.

Norman replied,
>Speeding up of 'minimize' is on my todo list already.  I'll put your email in my notes for that task, to take your suggestions into account when I work on it.  I think there is some simple constant pre-matching speedup already at the proof step level, but not globally for the proof as you suggest, which should be be faster. 

I still think this idea is worth pursuing, and I believe it should reduce minimization time dramatically. I haven't produced any working code unfortunately, besides this Haskell tool I mentioned in the letter, which doesn't do the minimization itself.

Issue #93, opened 2022-07-15
Formatting bug


benjub:
As noticed in https://github.com/metamath/set.mm/pull/2689#discussion_r921835561, it looks like in some cases (but not all) when a $p-statement is written with `$p` ending a line (hence the next line begins with `|-`), then the proof is not indented (see currently ~eucalg in set.mm, but notice that e.g. ~stirlinglem5's proof is correctly indented).

savask:
I think I found the root of the bug. Here's a small example:
```
  $c A $.
  $( An axiom. $)
  a1 $a A $.
  ${
    $( A theorem.  (Contributed by XX, 01-01-1970.) $)
    th $p
      A $= ( a1 ) A $.
  $}
```
After `save proof */compressed/fast` and  `write source ... /rewrap` we get
```
  $c A $.
  $( An axiom. $)
  a1 $a A $.
  ${
    $( A theorem.  (Contributed by XX, 01-01-1970.) $)
    th $p
      A $=
  ( a1 ) A $.
  $}
```
The problem seems to be in the [getSourceIndentation](https://github.com/metamath/metamath-exe/blob/adb9bec92afeb0dbc65fee0aa7c029bcf8c4fb8f/src/mmdata.c#L3106) function of the file `mmdata.c`. It is used to compute indentation for the proof block, so when it's called in our example the `fbPtr` pointer points at a newline right after $p token, and this immediately stops the `while(1)` loop computing indentation.

It seems that the fix is too add the line
```
  if (fbPtr[0] == '\n') fbPtr--;
```
right before the while loop. I ran the version of metamath with the fix on set.mm, and the only affected lines are the ones with proofs with broken indentation, so it seems that the fix doesn't do any harm. Formatting is a delicate subject, so I would be glad to hear opinions of other metamath-exe contributors (maybe @digama0 or @wlammen can take a look?).

wlammen:
I think one can unconditionally step back one character, so you are on the p of the $p (if this is the keyword), no matter what character follows it
The comment `/* Go back to first line feed prior to the label */` is misleading.  The first line feed (LF) before the key determines the indentation.  In theory the label and keyword may be separated by one or more LF.

benjub:
@wlammen or @savask : can you make a PR with the fix you proposed ?

wlammen:
I am not going to add a PR on my behalf, before #100 is finished, and that project seems to have grind to a standstill already at the second step.

savask:
[...]

PR #94, opened 2022-07-17
Improve minimize performance

savask:
This is an implementation of the idea presented in #10 . One precomputes the set of math constants used in the proof of the minimized statement (it is stored in .tmp field of g_MathToken in the code), and for each statement we check if all its constants lie in this set. Statements which pass this validation step are marked with 1 in the boolean array usefulStatements. Then the usual back-and-forth minimize_with loop is run, but now it checks only "useful statements".

If minimize_with is supplied with the /VERBOSE option, it will output the number of potentially useful theorems at the beginning.

This optimization reduces the minimization time twofold. For example, on my computer `smfsupmpt` requires 94s to minimize with the vanilla version, and 42s with this optimization; `ssmapsn`` requires 32s without and 13s with the optimization.

One downside of my code is the allocation of the usefulStatements array, I would prefer it to be an expandable vector (not a boolean array), but as far as I understand metamath codebase doesn't have a vector implementation.

### DIFF
--- a/src/metamath.c
+++ b/src/metamath.c
@@ -5150,6 +5150,92 @@ void command(int argc, char *argv[]) {
       print2("Bytes refer to compressed proof size, "
         "steps to uncompressed length.\n");
 
+      /* Clear tmp for all math tokens */
+      for (k = 0; k < g_mathTokens; k++)
+        g_MathToken[k].tmp = 0;
+
+      long step, mlen;
+      nmbrString *mString;
+
+      /* Set tmp to 1 for math tokens present in the proof */
+      m = nmbrLen(g_ProofInProgress.proof);
+      for (step = 0; step < m; step++) {
+        mString = (g_ProofInProgress.target)[step];
+        mlen = nmbrLen(mString);
+
+        for (i = 0; i < mlen; i++) {
+          g_MathToken[mString[i]].tmp = 1;
+        }
+      }
+
+      /* usefulStatements is a boolean array, where usefulStatements[k] == 1
+         iff statement k might be useful in minimization */
+      char *usefulStatements;
+      usefulStatements = poolMalloc(g_proveStatement*sizeof(char));
+      long usefulCount = 0;
+
+      /* Fill the usefulStatements array */
+      for (k = 1; k < g_proveStatement; k++) {
+        usefulStatements[k] = 1;
+
+        if (!mathboxFlag && k >= g_mathboxStmt && k < thisMathboxStartStmt) {
+          usefulStatements[k] = 0;
+          continue;
+        }
+
+        if (g_Statement[k].type != (char)p_ && g_Statement[k].type != (char)a_) {
+          usefulStatements[k] = 0;
+          continue;
+        }
+        if (!matchesList(g_Statement[k].labelName, g_fullArg[1], '*', '?')) {
+          usefulStatements[k] = 0;
+          continue;
+        }
+
+        if (exceptPos != 0) {
+          /* Skip any match to the EXCEPT argument */
+          if (matchesList(g_Statement[k].labelName, g_fullArg[exceptPos + 1],
+                '*', '?')) {
+            usefulStatements[k] = 0;
+            continue;
+          }
+        }
+
+        if (forbidMatchList[0]) { /* User provided a /FORBID list */
+          /* First, we check to make sure we're not trying a statement
+             in the forbidMatchList directly (traceProof() won't find
+             this) */
+          if (matchesList(g_Statement[k].labelName, forbidMatchList, '*', '?')) {
+            usefulStatements[k] = 0;
+            continue;
+          }
+        }
+
+        /* Check to see if statement comment specified a usage
+           restriction */
+        if (!overrideFlag) {
+          if (getMarkupFlag(k, USAGE_DISCOURAGED)) {
+            usefulStatements[k] = 0;
+            continue;
+          }
+        }
+
+        /* Check if all math constants used in statement k are present
+           in the minimized proof. Hypotheses are not checked. */
+        mString = g_Statement[k].mathString;
+        mlen = nmbrLen(mString);
+
+        for (i = 0; i < mlen; i++) {
+          if (g_MathToken[mString[i]].tokenType == (char)con_) {
+            if (!g_MathToken[mString[i]].tmp)
+              usefulStatements[k] = 0;
+          }
+        }
+        usefulCount += usefulStatements[k];
+      }
+      if (verboseMode)
+        print2("Found %ld potentially useful theorems.\n", usefulCount);
+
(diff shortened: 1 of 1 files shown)


benjub:
I'm not sure I understand. It may be the case that a proof is shortened by statements using constants not appearing in the current proof. Maybe you know in advance that such shortenings will not be caught by the current minimize_with program, so you're saying we might as well discard them from the beginning ?

savask:
>Maybe you know in advance that such shortenings will not be caught by the current minimize_with program, so you're saying we might as well discard them from the beginning ?

Yes, minimize_with X tries to use X at one of the steps of the proof, so naturally all constants used in X must be already present in the proof of our statement. I can imagine a multi-step subproof using some new constants yet shortening the proof of the original statement, but minimize_with can't find those, unfortunately.

src/metamath.c
```diff
@@ -5150,6 +5150,92 @@ void command(int argc, char *argv[]) {
...
+
+      /* usefulStatements is a boolean array, where usefulStatements[k] == 1
+         iff statement k might be useful in minimization */
+      char *usefulStatements;
+      usefulStatements = poolMalloc(g_proveStatement*sizeof(char));
+      long usefulCount = 0;
+
+      /* Fill the usefulStatements array */
+      for (k = 1; k < g_proveStatement; k++) {
+        usefulStatements[k] = 1;
```
wlammen, line 5179: I suggest to preset it to 0, so you dont need all the usefulStatements[k] = 0 instructions for the early-out-of-loopbody instructions.

src/metamath.c
```diff
@@ -5150,6 +5150,92 @@ void command(int argc, char *argv[]) {
...
+
+        /* Check if all math constants used in statement k are present
+           in the minimized proof. Hypotheses are not checked. */
+        mString = g_Statement[k].mathString;
+        mlen = nmbrLen(mString);
+
+        for (i = 0; i < mlen; i++) {
+          if (g_MathToken[mString[i]].tokenType == (char)con_) {
+            if (!g_MathToken[mString[i]].tmp)
+              usefulStatements[k] = 0;
```
wlammen, line 5231: This should be followed by a break; instruction

wlammen (changes requested):

david-a-wheeler:
> metamath codebase doesn't have a vector implementation.

It's certainly possible to add one. The usual approach is a simple data structure with a count of the "number of elements" and a pointer to the (beginning of the) array elements; if not NULL it needs have been allocated (e.g., with malloc). To resize, call realloc() to resize the array, move things as necessary, and change the count. I did a quick Google search and found this one, implemented as OSS: https://github.com/eteran/c-vector

wlammen:
> https://github.com/eteran/c-vector
And it is C89 (ANSI C https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf) compatible.  I wonder whether we should give up on this and at least advance to C11.

david-a-wheeler:
> I wonder whether we should give up on this and at least advance to C11.

We could move to C11 if that's helpful. The key is compiler support in deployed systems. [GCC seems to have good C11 support](https://gcc.gnu.org/wiki/C11Status#:~:text=GCC%204.9%20Changes%3A%20%E2%80%9CISO%20C11,optional%20C99%20features%20from%20Annexes).

However, I don't think that helps. C11 doesn't have a vector type as far as I know, and <https://github.com/eteran/c-vector> doesn't require C11. We should write code that is *compatible* with C11, but I think we should only require C11 if there's advantage to it. I didn't see any advantages in [this list of C11 additions](https://en.wikipedia.org/wiki/C11_%28C_standard_revision%29)

wlammen:
> However, I don't think that helps.

See for example https://en.cppreference.com/w/c/io/fprintf to see how much C has developed since 1989.  Functions like snprintf really make sense. Or // style comments.

`gcc -O3 -funroll-loops -finline-functions -fomit-frame-pointer -DINLINE=inline -g  -Wall -Wextra   -o metamath metamath.o mmcmdl.o mmcmds.o mmdata.o mmfatl.o mmhlpa.o mmhlpb.o mminou.o mmpars.o mmpfas.o mmunif.o mmveri.o mmvstr.o mmword.o mmwtex.o`

This is the command executed by the build.sh script.  It does not check for ANSI C.  The option `-ansi` should be added to enforce this standard.  By the way, this can be done by changing line 4567 to new_CFLAGS="-Wall -Wextra -ansi".

david-a-wheeler:
I agree snprintf and // are useful, but I those are C99 additions, not C11.

I'm not opposed to switching to a later C spec, I just want to make sure that a spec move is justified *and* that it is widely implemented by commonly-available & distributed compilers.

We should at *least* move up to C99, no question in my mind!!! C99 is very widely supported & I think we're already using its facilities. The only question is if we should move to C11 or not.

digama0:
We are already on C11. The build script uses it and `//` can be spotted in the source code.

savask:
>It's certainly possible to add one.

It is, but is it worth doing so for one application? Also, metamath-exe uses its own memory management system, for example, in code I used `poolMalloc` instead of `malloc` and `poolFree` instead of `free`. I was wondering if `nmbrString` can be used as vector of numbers, but I haven't been able to find a `push_back` equivalent in mmdata.c.

wlammen:
> We are already on C99.

Indeed: CONTRIBUTIONS.md states this literally.

src/metamath.c
savask: Done.

src/metamath.c
```diff
@@ -5150,6 +5150,86 @@ void command(int argc, char *argv[]) {
...
+
+        for (i = 0; i < mlen; i++) {
+          g_MathToken[mString[i]].tmp = 1;
+        }
+      }
+
+      /* usefulStatements is a boolean array, where usefulStatements[k] == 1
+         iff statement k might be useful in minimization */
+      char *usefulStatements;
+      usefulStatements = poolMalloc(g_proveStatement*sizeof(char));
```
benjub, line 5174: Why not use `bool`, `true`, `false` as advised in https://github.com/metamath/metamath-exe/issues/17#issue-1091159614 ? Then a few statements below have minor modifications, e.g., `usefulCount += usefulStatements[k];` has to be replaced by something like `if (usefulStatements[k]) {usefulCount += 1}`

benjub:
Having @wlammen's approval is already a pretty good guarentee ! If @digama0 can find some time to review and if ok merge this, this would be nice and this would also unblock #93.

src/metamath.c
savask: I think originally I used `char` for consistency, for example, the `.tmp` field uses `0` and `1` instead of `bool`. As a positive side effect, `usefulCount` can be updated without branching (although I don't think it makes any great difference in this case).

PR #101, opened 2022-09-28
fatal error handling #2

wlammen:
Here we add a framework for regression tests.

Two macros control whether regression tests are compiled in:
1. **REGRESSION_TEST** is set from the outside.  There are several ways to set this macro, but the preferred one is calling build.sh with the -t option.  This is not always sufficient, because sources need to change to trigger a compile.  It is safest to use `./build.sh -ct` to enable tests, and `./build.sh -c` to disable them again.
2. In the header mmfatl.h the macro **TEST_MMFATL** can be unconditionally un/defined to override the settings in 1.

After a successful compile invoking the executable ./metamath will run the test suite right on program start.  So far no tests are included, so the run always displays a successful completion.  The inclusion/absence of this message is a visual indication of an execution of regression tests and can be used to verify the expected behavior.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -10,9 +10,7 @@ help_text=\
 'Builds all artefacts in the metamath-exe/build subfolder (if not directed
 otherwise).  Change to the metamath-exe top folder first before running
 this script, or issue the -m option.
-
 Possible options are:
-
 -a Used by autotools. Put hyphens in the version string when used with -v
 -b build binary only, no reconfigure. Faster, but should not be used on first run.
 -c Clean the build directory.
@@ -22,6 +20,7 @@ Possible options are:
     Relative paths are relative to the current directory.
 -o followed by a directory: optionally clean directory and build all artefacts there.
     Relative paths are relative to the destination'"'"'s top metamath-exe directory.
+-t run regression tests in the executable
 -v extract the version from metamath sources, print it and exit'
 
 #============   evaluate command line parameters   ==========
@@ -34,8 +33,9 @@ version_only=0
 version_for_autoconf=0
 unset dest_dir
 top_dir="$(pwd)"
+do_tests=0
 
-while getopts abcdhm:o:v flag
+while getopts abcdhm:o:tv flag
 do
   case "${flag}" in
     a) version_for_autoconf=1;;
@@ -45,6 +45,7 @@ do
     h) print_help=1;;
     m) cd "${OPTARG}" && top_dir=$(pwd);;
     o) dest_dir=${OPTARG};;
+    t) test_flag="CFLAGS=-DREGRESSION_TEST";;
     v) version_only=1;;
   esac
 done
@@ -116,7 +117,7 @@ then
   autoreconf -i
 
   cd "$build_dir"
-  "$top_dir/configure" -q
+  "$top_dir/configure" -q $test_flag
 fi
 
 #===========   do the build   =====================
--- a/build.sh
+++ b/build.sh
@@ -10,7 +10,9 @@ help_text=\
 'Builds all artefacts in the metamath-exe/build subfolder (if not directed
 otherwise).  Change to the metamath-exe top folder first before running
 this script, or issue the -m option.
+
 Possible options are:
+
 -a Used by autotools. Put hyphens in the version string when used with -v
 -b build binary only, no reconfigure. Faster, but should not be used on first run.
 -c Clean the build directory.
--- a/src/mmfatl.c
+++ b/src/mmfatl.c
@@ -10,4 +10,28 @@
  * conditions (corrupt state, out of memory).
  */
 
+#include <stdbool.h>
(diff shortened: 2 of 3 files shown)


wlammen:
Ready for review

wlammen:
Ready for review

wlammen:
@digama0 Is there any reason to keep this further on hold?

digama0:
Sorry, as I mentioned before I don't have much time to spend on reviewing for metamath-exe. Please ping me if I'm not responsive. The PR looks good; the only change I would suggest is to modify the CI script to run the regression tests.

wlammen:
I don't want to be too far distracted from the main purpose of this PR series, which is providing a separate software layer for fatal errors, in particular under severe memory restrictions.  This renders the software more standard compatible as it is now.  CI integration is a different topic, and I actually have no clue how the scripts interacts with programs.  I will change the code such, that it terminates the test suite with exit code 0 and no output on success, else with code 1 (or what is prescribed in case of failure) and an error message to stderr.  If this is not enough, a different PR should care about anything more.

digama0:
The point of CI is to test each new addition so that we can be reasonably confident in the correctness of the new code. It is supposed to run *on the PR that adds the functionality*, so if you are adding a new test suite then CI for that test suite is part of the PR.

wlammen:
I need a reference to the expected behaviour in CI scripts.

digama0:
If regression tests fail, then the CI should fail, meaning that things show up in red in the github interface. That's the main thing. Using exit codes is the easiest way to get the C code to communicate success / failure to the CI script.

When tests fail, the logging should also be reasonably descriptive about what failed, and conversely the logging shouldn't be too verbose when everything is working well. The regular test suite is an example of a fairly full featured test harness, we probably don't need all of those behaviors but you can use it for inspiration.

david-a-wheeler:
I agree with Mario on this. As far as exit codes go, "0" means "no errors" (aka true) and any other value means an error (aka false). If a CI test fails, it needs to return nonzero. 1 is fine, but it doesn't have to be 1. We want the CI test suite to run & report errors if there's an error, that will prevent many problems long term.

wlammen:
The correct exit code to use is EXIT_FAILURE or EXIT_SUCCESS (https://en.cppreference.com/w/c/program/EXIT_status), which often are 1 and 0, but can be different.
I provided a test suite with my PR that can be manually activated.  It is somewhat out of the scope of this task, though, to integrate regression tests in Metamath generally with Github CI, in particular because I am not familiar with this framework.  Let's see.  If it does not take too much time, I'll do it.

digama0:
> In order to fix the final commit, the build.yml has to be adapted. I do not have sufficient rights.

What do you mean by this? It's just a regular file in the repo, you should be able to commit changes to it.

wlammen:
The presence of builtin.in allows some fine tuning.  If you remove builtin.in from tests, the regression tests are not executed.  Don't know whether this is an advantage or not.

digama0:
I was actually thinking that the regression tests would be a separate job: rather than building both executables at once there is one job that runs the integration tests and another that runs the regression tests, so the github interface can distinguish the two cases. (Currently it says "1 successful check" because there is only one job; if you look at set.mm you will see "8 successful tests" because there are 8 separate jobs running all the different verifiers.

wlammen:
I know there is headroom for optimization.  But maybe that is really a different PR.

build.sh
```diff
@@ -121,6 +124,15 @@ fi
...
+then
+  # create an executable running regression tests
+  make "CFLAGS=-DBUILD_REQUESTS_REGRESSION_TEST"
+  mv src/metamath "$top_dir"/metamath_test
+  make clean
+fi
+
+# normal executable
 make
 mv src/metamath "$top_dir"
```
digama0, line 137:
```suggestion
else
  # normal executable
  make
  mv src/metamath "$top_dir"
fi
```

digama0:
I'm not really talking about optimization? Could you remove the changes to `run_tests.sh`, since they aren't needed this way.

wlammen:
Its almost midnight now.  I need to stop here.

wlammen:
Won't do more

wlammen:
Note this warning from actions/checkout@v2

Warning: The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

wlammen:
I'd rather stop here.  Regression tests are up and running, there is some support already to migrate code to other modules.  I am pretty sure those who have already developed test systems can come up with better ideas.  It is not my primary goal in this PR, to provide a perfect support for regression tests, but provide a platform that allows my own tests being run.
@digama0 ready to merge, so I can proceed?

wlammen:
@digama0 thanks

PR #102, opened 2022-10-14
fatal error handling metamath:3

wlammen:
The layer printing a final fatal error message must not rely on the memory heap, as it can be exhausted, or even corrupt.  Instead it uses a buffer allocated right at program start for this sole purpose.  It needs this buffer, because it is intended to add simple parsing features to the fatal error message handler.

2 macros control the details of this buffer:
- BUFFERSIZE the capacity of the buffer.  The terminating NUL character must be covered
- ELLIPSIS a string indicating truncation to the user should the parsing overflow the buffer.

This PR allocates a buffer according to the macro settings.  Although C initializes all global variables, we do not rely on this, but assume the buffer might be overwritten in a corrupt environment, and we need to initialize it again.  So far this initialization is very simple: Fill the buffer with NUL and append the ELLIPSIS at the end.

This functionality can undergo a regression test.  So this PR adds the first entry to the test suite.  A few adaptions were made to the macros controlling whether a normal executable, or a test version is to be created.  Boilerplate code for a more concise test code added.

### DIFF
--- a/src/metamath.c
+++ b/src/metamath.c
@@ -703,8 +703,6 @@
 
 /*----------------------------------------------------------------------*/
 
-#undef RUN_REGRESSION_TEST
-
 #include <string.h>
 #include <stdlib.h>
 #include "mmvstr.h"
@@ -746,7 +744,6 @@ int main(int argc, char *argv[]) {
 #ifdef TEST_ENABLE // enable this in mmtest.h or via './build.sh -t'
 
   RUN_TESTS();
-  exit(EXIT_SUCCESS);
 #endif
 
   /****** If g_listMode is set to 1 here, the startup will be Text Tools
--- a/src/metamath.c
+++ b/src/metamath.c
@@ -703,6 +703,7 @@
 
 /*----------------------------------------------------------------------*/
 
+#undef RUN_REGRESSION_TEST
 
 #include <string.h>
 #include <stdlib.h>
@@ -742,7 +743,6 @@ int main(int argc, char *argv[]) {
 /* argc is the number of arguments; argv points to an array containing them */
 
 # ifdef RUN_REGRESSION_TEST
-
 // activated if at least one unit requests regression tests
 
   bool regressionTestResult = true;
--- a/src/metamath.c
+++ b/src/metamath.c
@@ -742,8 +742,8 @@ int main(int argc, char *argv[]) {
 /* argc is the number of arguments; argv points to an array containing them */
 
 #ifdef TEST_ENABLE // enable this in mmtest.h or via './build.sh -t'
-
   RUN_TESTS();
+  // you never get here
 #endif
 
   /****** If g_listMode is set to 1 here, the startup will be Text Tools
--- a/src/metamath.c
+++ b/src/metamath.c
@@ -743,13 +743,12 @@ int main(int argc, char *argv[]) {
 /* argc is the number of arguments; argv points to an array containing them */
 
 # ifdef RUN_REGRESSION_TEST
-// activated if at least one unit requests regression tests
+// activated if at least one unit requests regression tests.
+// a failing test MUST terminate the program with exit(EXIT_FAILURE).
 
-  bool regressionTestResult = true;
+  test_mmfatl();
 
-  test_mmfatl(&regressionTestResult);
-
-  exit(regressionTestResult? EXIT_SUCCESS : EXIT_FAILURE);
+  exit(EXIT_SUCCESS);
 
 # endif
 
--- a/src/metamath.c
+++ b/src/metamath.c
@@ -721,6 +721,7 @@
 #include "mmword.h"
 #include "mmwtex.h"
 #include "mmfatl.h"
+#include "mmtest.h"
 
 void command(int argc, char *argv[]);
 
@@ -742,15 +743,11 @@ int main(int argc, char *argv[]) {
 
 /* argc is the number of arguments; argv points to an array containing them */
 
-# ifdef RUN_REGRESSION_TEST
-// activated if at least one unit requests regression tests.
-// a failing test MUST terminate the program with exit(EXIT_FAILURE).
+#ifdef TEST_ENABLE // enable this in mmtest.h or via './build.sh -t'
 
-  test_mmfatl();
-
+  RUN_TESTS();
   exit(EXIT_SUCCESS);
-
-# endif
+#endif
 
   /****** If g_listMode is set to 1 here, the startup will be Text Tools
           utilities, and Metamath will be disabled ***************************/
--- a/src/mmfatl.c
+++ b/src/mmfatl.c
@@ -72,20 +72,20 @@ static void initBuffer()
 
 /*
  * Use this extension of CHECK_TRUE if file and line number is not
- * sufficient to locate the failing test.  context is a string constant. 
+ * sufficient to locate the failing test.  'context' is a string constant. 
  */
-#define CHECK_TRUE_W_CONTEXT(bool_expr, context)         \
-    if(!(bool_expr)) {                                   \
-        printf(context);                                 \
-        printf("assertion %s failed", #bool_expr);       \
+#define CHECK_TRUE_W_CONTEXT(bool_expr, context)   \
+    if(!(bool_expr)) {                             \
+        printf("%s:\n", context);                  \
+        printf("assertion %s failed", #bool_expr); \
         printf(" at %s:%u\n", __FILE__, __LINE__); \
-        exit(EXIT_FAILURE);                              \
+        exit(EXIT_FAILURE);                        \
     }
 
 /*
  * If bool_expr evaluates to false, print the message and return to caller
  */
-#define CHECK_TRUE(bool_expr) CHECK_TRUE_W_CONTEXT(bool_expr, __func__" ")
+#define CHECK_TRUE(bool_expr) CHECK_TRUE_W_CONTEXT(bool_expr, __func__)
 
 #define REGRESSION_TEST_SUCCESS            \
     printf("Regression tests in " __FILE__ \
--- a/src/mmfatl.c
+++ b/src/mmfatl.c
@@ -18,25 +18,11 @@
 
 /*!
  * character constant
- */ 
+ */
 #define NUL '\x00'
 
 //----------
 
-/*
- * During development you may not want to expose preliminary results to the
- * normal compile, as this would trigger 'unused' warnings, for example.  In
- * the regression test environment your code may be referenced by testing code,
- * though.
- *
- * This section should be empty, or even removed, once your development is
- * finished.
- */
-#if TEST_ENABLE
-#   define UNDER_DEVELOPMENT
-#endif
-
-#ifdef UNDER_DEVELOPMENT
 /* the size a fatal error message including the terminating NUL character can
  * assume without truncation.
  */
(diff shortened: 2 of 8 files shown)


wlammen:
@digama0 ready for review

src/mmfatl.h
```diff
@@ -79,54 +79,58 @@
...
 
 /* -----   copy & paste code, the same in all modules -----  */
 
-/* in case an included header file of another module has defined it */
-#undef TEST_ENABLE
-
 /* If BUILD_REQUESTS_REGRESSION_TEST is defined compilation of regression
  * tests is requested from somewhere outside of C files.  This can still be
- * overridden locally by setting TEST_ENABLE explicitely below.
+ * overridden locally by setting TEST_FORCE_ENABLE explicitely before.
```
digama0, line 97:
```suggestion
 * overridden locally by setting TEST_FORCE_ENABLE explicitly before.
```

src/mmfatl.c
```diff
@@ -34,7 +125,8 @@ void test_mmfatl(bool* ok)
     bool silent = TEST_SILENT;
 
     *ok = *ok
+        && test_initBuffer(silent)
```
digama0, line 128: Can you have each successful test log that it was successful in CI? We don't need the details of the test but it's important to know that the test itself ran (compare with the regular test suite).

wlammen: This will in the end be awfully verbose, because my initial PR #86 already had 30+ individual tests.  I don't recommend this.  But one message for each module is ok, I think

src/mmfatl.c
```diff
@@ -10,18 +10,109 @@
...
+    if (ok)
+    {
+        ok = buffer[i] == NUL;
+        if (!ok)
+            printf("initBuffer: "
+                "no NUL character following the ellipsis at the end\n");
+    }
+    if (ok && !silent)
+        printf("initBuffer OK\n");
+    return ok;
```
digama0, line 113: I'm not very happy about how much code this test is. I have less confidence in the test than in the original code. The detailed printf errors are not necessary, and without them the code gets a lot simpler:
```
for (unsigned i = 0; i < BUFFERSIZE; ++i)
  if (buffer[i] != NUL) return false;
for (unsigned i = 0; i < sizeof(ELLIPSIS); ++i)
  if (buffer[BUFFERSIZE + i] != ELLIPSIS[j]) return false;
return true;

File: src/mmfatl.c (hunk 2 above)
Comment (wlammen) [reply src/mmfatl.c:130]: In case of suspicion you can still checkout the sources, uncomment the # define TEST_FORCE_ENABLE_LINE and compile it the usual way with ./build.sh -c .  This will run the regression test with all the gory  details.

File: src/mmfatl.c (hunk 3 above)
Comment (wlammen) [reply src/mmfatl.c:113]: ok.  So just a summary, and letting the programmer find out the details by hand.  Fine with me.

File: src/mmfatl.c (hunk 2 above)
Comment (digama0) [reply src/mmfatl.c:130]: > This will in the end be awfully verbose, because my initial PR #86 already had 30+ individual tests. I don't recommend this. But one message for each module is ok, I think

Yes, I expect to see 30 lines in CI. This is pretty standard for testing frameworks. 30 lines is not a lot, especially if you use some colors to make things stand out, or use very similar formatting for each test line so that a failing test looks very different from the surrounding code.

File: src/mmfatl.c (hunk 3 above)
Comment (wlammen) [reply src/mmfatl.c:92]: > I have less confidence in the test than in the original code.

If you program a task using two different methods, getting the same results, the confidence is to some extent achieved, even if each method is complex.

File: src/mmfatl.c [hunk 4]
```diff
@@ -10,22 +10,94 @@
...
+    // ...and has the ELLIPSIS string at the end...
+    char ellipsis[] = ELLIPSIS;
+    unsigned j = 0;
+    for (; ok && ellipsis[j] != NUL; ++i, ++j)
+        ok = buffer[i] == ellipsis[j];
+    ok = ok && buffer[i] == NUL;
+    if (!ok)
+        printf("initBuffer: initialization failed\n");
+    else if (!silent)
+        printf("initBuffer OK\n");
```
digama0, line 91: This line (and the logic around `silent`) should be in the test harness, not the actual test. I think the test should only be responsible for the failure message and boolean return value, and the harness follows that up with general information about the tests. (You can use a macro to stringify the test name to put it in the printf.) So the harness would look something like:
```
RUN_TEST(test_initBuffer);
RUN_TEST(testSuccessMessage);
```
and there is some code around this and in the macro to handle accumulating an `ok` variable and printing ok/fail stuff.

wlammen: https://en.wikipedia.org/wiki/List_of_unit_testing_frameworks#C

One problem is that C99 predates many of them...
Lets address this step by step.  Again, it is not my primary goal in this series to present a fool-proof test suite, just to give it a head start.

src/mmfatl.c
wlammen: 30 lines for each module (i.e mmxxxx.c file) sums up to several 100

src/mmfatl.c
digama0: I wasn't suggesting using an off the shelf unit testing framework (although the way this is going I wouldn't be surprised if it ends up being a better investment in the end). Here `RUN_TEST(x)` is something like:
```c
#define RUN_TEST(x) \
  if (x()) { printf("test " #x " ok\n"); } else { ok = false; printf("TEST " #x " FAILED\n"); }
```

digama0: For now I'm trying to keep the code growth as a result of these tests as small as possible, so low-tech and concise is good. We still aren't testing anything that really matters here, like verification and proof assistant activities, so this investment hasn't paid off yet from my perspective.

wlammen: ok, tomorrow.  Its about midnight here.

src/mmfatl.c
wlammen: I think I'll make it configurable.

src/mmfatl.c
wlammen: 1. It is easier to decide where to refactor and optimize when you have a handful test cases at hand.  It is a bit more difficult to anticipate this in advance.  Or in other words, the code gets better with the experience you have.
2. I caught two bugs with the parser in my original #86 PR.  And I needed to convert the source from int to bool, and the correctness was verified by the test suite.  Not bad, I'd say.

wlammen:
@digama0 A bit inspired by some unit tests from the list posted above, I added boilerplate code for checks.  This now slims down the test code considerably.
In commit [32a703a](https://github.com/metamath/metamath-exe/pull/102/commits/32a703ac32b679f50b5b0572cb27f31c3b28c0fa) I deliberately added a failing test, so you can see how a detected regression looks like in github checks.
In [96cfda3](https://github.com/metamath/metamath-exe/pull/102/commits/96cfda373e77ae8acd366eaedd4735821ea0d324) I forced TEST_SILENT to false, so each individual test prints an OK line now.  This meets one of your requests.

wlammen:
@digama0 ?

digama0:
I'll look at this tonight

src/metamath.c
```diff
@@ -742,14 +743,12 @@ int main(int argc, char *argv[]) {
 /* argc is the number of arguments; argv points to an array containing them */
 
 # ifdef RUN_REGRESSION_TEST
+// activated if at least one unit requests regression tests.
+// a failing test MUST terminate the program with exit(EXIT_FAILURE).
```
digama0, line 747: If a test terminates the program, how will the rest of the tests run? I think it should instead either throw an exception (not an option in C AFAIK) or return failure so that the test harness can aggregate information on all successes and failures.

wlammen: If your modifications, say, broke a primitive routine, then all tests based on this primitive are likely to fail as well.  It is pointless to watch such a cascading effect in total.  Compilers follow a similar strategy.  A missing brace may entail hundreds of grammar errors in the following code lines, but usually only the first 10 or so are shown, before the process stops.

It is a matter of taste, or, if you like, minor improvement, to not stop at the first error, but continue up to the n-th one.  I don't mind if you, or someone else, modifies the process in that direction.  But please, do not expect me to do this optimization.

See, after about 3 weeks this PR contains the allocation and initialization of a buffer.  That's it.  About 6 lines of my own code wait to pass the review, because I am quite some time bothered with creating an infrastructure, that wasn't in place before, and that, strictly speaking, is not necessary for my primary goal.  We need to make some progress on the core ideas as well.

digama0: You previously had an implementation that would do exactly this (with the `ok` variable that was being assigned in the harness), I'm not sure why you removed it.

It's true that tests can often fail in batches. But not all bugs are like that, and if the test suite has good coverage then hopefully they will be more orthogonal than that. I think the simpler thing is just to let them all run and let the logs have the information necessary.

You seem to be frustrated that it is taking a while to add this test suite, but I believe the idea of a regression test suite was your idea. If you are going to do a test suite, please take the time to make it good, or don't bother in the first place and use the test suite we already have.

digama0:
The assertion failures look good. If you have a header before each test we can also match them to individual tests. (Why do we need to know what test they are in if the assertions have a file+line, you say? Because the assertions might be in common test code or even in the main metamath functions, so we would only know what function is failing but not what test caused the failure.)

wlammen:
> Because the assertions might be in common test code or even in the main metamath functions

These are assumptions about how testing could develop in future, should it be used on a larger scale.  They may hold or not, but in any case, they are not connected to what this PR is initially about.  That's a problem, because I cannot get my work done.  If I am unlucky I wait again for several days before this "extension" is accepted.

I understand that this is a learning curve.  So I think I should voice my position, too.  It is a pain to be constantly distracted from my goals, and get something done that is best addressed in a different task.  As a minimum, if you still think I should fix some minor, non-critical issues, though, could you perform a merge, and then ask whether I got some time to add this or that in the next PR?

src/metamath.c
wlammen: I added my personal hand-written ad-hoc code in #86 to ease maintenance of this piece of software.  For one I wanted to make sure I got everything right.  And then I wanted to be prepared if people request changes.  The test code was actually used in both cases, and proved useful.  It was hidden behind a ifdef .. endif clause and was so never part of a regular build.  It could simply have been ignored.

My expectations was that after some initial discussions modifications peter out and the software won't be touched for years again.  In this phase the test code could have been safely dropped.

In hindsight I should have kept the test code on my computer and simply included it in my private build each time someone plays around with the code.  And complain if the tests fail.

I am not opposed to provide people with a regression test infrastructure.  I think it is a MUST for mathematical software like this, because it becomes part of the reasoning and replaces peer review.  So confidence is an issue.  And my review on pending PR from others show me how important this is to asses contributions of others.

But I am opposed to work on a test framework beyond the scope needed for this PR.  By the way, I think another infrastructure layer supporting debugging in a systematic way is necessary, too (Won't open this can of worms, though).

digama0:
Okay, since you are in a hurry to get this merged I've pushed all the changes I'm suggesting to your branch. If you are okay with it I will merge.

src/metamath.c
wlammen: > I'm not sure why you removed it

Implementations of outside test frameworks suggested this, and I followed the pattern.  One obvious result is the code lines reduce, because all is done in a macro call.

wlammen:
Thank you

src/metamath.c
```diff
@@ -703,6 +703,7 @@
 
 /*----------------------------------------------------------------------*/
 
+#undef RUN_REGRESSION_TEST
```
wlammen, line 706: can be deleted

src/metamath.c
```diff
@@ -741,17 +743,11 @@ int main(int argc, char *argv[]) {
...
-
-  bool regressionTestResult = true;
-
-  test_mmfatl(&regressionTestResult);
-
-  exit(regressionTestResult? EXIT_SUCCESS : EXIT_FAILURE);
-
-# endif
+  RUN_TESTS();
+  exit(EXIT_SUCCESS);
```
wlammen, line 749: RUN_TESTS exits already

src/mmtest.h
```diff
@@ -0,0 +1,101 @@
...
+ * the context of the failure.
+ *
+ * We recommend running the tests each time you modify the code to ensure it
+ * still executes as desired.  They are automatically invoked by Github checks
+ * on each push request.
+ */
+
+/*!
+ * \def TEST_ENABLE
+ * macro, is either 0 or 1.
```
wlammen, line 40: This is not correct

src/mmtest.h
```diff
@@ -0,0 +1,101 @@
...
+
+  /*
+  * If bool_expr evaluates to false, print an error message and return to
+  * caller.
+  *
+  * Use this extension of ASSERT if file and line number is not
+  * sufficient to locate the failing test.
+  * Accepts a format string for the assertion message.
+  */
+  #define ASSERTF(bool_expr, ...)                \
```
wlammen, line 76: I wasn't sure about C99 and variadic macros, so I avoided them.
https://stackoverflow.com/questions/3576396/variadic-macros-with-0-arguments-in-c99

But ASSERT is a replacement for this corner case

wlammen:
The review is non-blocking

digama0:
feel free to apply fixes directly instead of leaving reviews. I agree with all your points (although I don't think the one about variadic macros applies; can you reproduce a warning there? Otherwise I think we should just not worry about it. It's not being used for 0-ary macros anyway.)

wlammen:
ready for merge

PR #103, opened 2022-10-19
fatal error handling part 4

wlammen:
Adding stream semantics to the buffer.

The parsing algorithm will scan through the format string and continuously add pieces of text to the output buffer.  It will use appendText() to perform this.  This function is meant to copy chunks of text from a source verbatim to the buffer.  It will stop only at a terminating NUL, a placeholder character %, or when the buffer overflows.  The source could either be the format string itself, or a parameter substituting a %s.  Both cases differ slightly:  In case of a parameter the % is just an ordinary character.  Which semantics to select is given by an argument TextType submitted to appendText.

### DIFF
--- a/src/mmfatl.c
+++ b/src/mmfatl.c
@@ -37,7 +37,7 @@
  * This section should be empty, or even removed, once your development is
  * finished.
  */
-#if TEST_ENABLE
+#ifdef TEST_ENABLE
 #   define UNDER_DEVELOPMENT
 #endif
 
@@ -120,8 +120,7 @@ static void initBuffer(void) {
  * used to indicate whether \ref PLACEHOLDER_CHAR is a normal character, or
  * is an escape character in a format string.
  */
-enum TextType
-{
+enum TextType {
   STRING, //<! NUL terminated text
   FORMAT  //<! NUL terminated format containing placeholders
 };
@@ -135,12 +134,10 @@ enum TextType
  *   insertion point of other data, and stops the copy process.
  * \return a pointer to the character following the last one copied.
  */
-static char const* appendText(char const* source, enum TextType type)
-{
+static char const* appendText(char const* source, enum TextType type) {
   char escape = type == FORMAT ? PLACEHOLDER_CHAR : NUL;
-  while (buffer.begin != buffer.end
-      && *source != NUL && *source != escape)
-        *buffer.begin++ = *source++;
+  while (buffer.begin != buffer.end && *source != NUL && *source != escape)
+    *buffer.begin++ = *source++;
   return source;
 }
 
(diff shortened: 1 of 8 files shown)


src/mmfatl.c
```diff
@@ -71,29 +153,110 @@ static void initBuffer(void) {
 
 bool test_initBuffer(void) {
   // emulate memory corruption
-  memset(buffer, 'x', sizeof(buffer));
+  memset(&buffer, 'x', sizeof(buffer));
 
   initBuffer();
 
+  ASSERT(buffer.begin == buffer.text)
```
digama0, line 160:
```suggestion
  ASSERT(buffer.begin == buffer.text);
```
Use semicolon after statement-like macros

src/mmfatl.c
```diff
@@ -71,29 +153,110 @@ static void initBuffer(void) {
...
+  ASSERT(buffer.text[i] == NUL);
+  return true;
+}
+
+char const* bufferCompare(char const* match, int from, unsigned lg,
+                        unsigned begin)
+{
+  if (memcmp(buffer.begin + from, match, lg) != 0)
+    return "unexpected buffer contents";
+  return buffer.begin == buffer.text + begin?
```
digama0, line 185:
```suggestion
  return buffer.begin == buffer.text + begin ?
```

src/mmfatl.c
```diff
@@ -71,29 +153,110 @@ static void initBuffer(void) {
...
+    (char const*)0 : "unexpected buffer begin";
```
digama0, line 186:
```suggestion
    (char const*)NULL : "unexpected buffer begin";
```
double check what the preferred way to write nullptr in C is

src/mmfatl.c
```diff
@@ -71,29 +153,110 @@ static void initBuffer(void) {
...
+ * \param from ... counting from this offset from buffer.begin after copy...
+ * \param lg ...and this many characters.
+ * \param begin offset of buffer.begin from buffer.text after copy.
+ * \return NULL on success, otherwise a message describing a failure
+ */
+char const* testcase_appendText(char const* text, unsigned adv,
+        char const* match, int from, int lg, unsigned begin)
+{
+    enum TextType type = *text == PLACEHOLDER_CHAR? FORMAT : STRING;
+    return appendText(text + 1, type) == text + adv + 1?
```
digama0, line 204:
```suggestion
    enum TextType type = *text == PLACEHOLDER_CHAR ? FORMAT : STRING;
    return appendText(text + 1, type) == text + adv + 0 1?
```

digama0 (approved):

wlammen:
@digama0 ready to merge

wlammen:
@digama0 I have no rights to merge this pull request despite of your approval.  The additions are merely cosmetic, and do not change behaviour.  Is it possible to finish this PR?  The next batch is waiting already.  It takes care of the other main aspect of the parser, the substitution of placeholders.  Further following is the integration of both aspects in a single parser class, after that its integration with error handling in Metamath proper.
I tried to make each PR as small as sensible, to ease review.

digama0:
You can queue up PRs, you know. Just submit the second PR with a note in the description like:

* [ ] depends on #102

You can do this whenever you have a sequence of dependent PRs, although it is preferable to make them independent if you can so that reviewing can happen in parallel.

PR #104, opened 2022-10-24
fatal error handling, part 5

wlammen:
@digama0 ready to review and merge.  This PR adds a function handleSubstitution that later will be invoked when the parser encounters a % character during his scan of the format string.  The purpose of this function is to check the placeholder type, fetch a parameter from a va_arg list and write the parameter to the output buffer, so that it appears to be inserted.  Afterwards the scan pointer is updated to point behind the placeholder.  There are a couple of irregular situations that are also handled: a % not followed by a recognized type character, overflowing output buffer, and a NULL pointer as a parameter to a string placeholder.

### DIFF
--- a/src/mmfatl.c
+++ b/src/mmfatl.c
@@ -13,6 +13,7 @@
 #include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <stdarg.h>
 #include <string.h>
 
 #include "mmfatl.h"
@@ -86,6 +87,8 @@
  * (see https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf).  We
  * support unsigned long in formatted error output to allow for macros like
  * __LINE__ denoting error positions in text files.
+ *
+ * There exist no utoa in the C99 standard library, that could be used instead.
  * \param value an unsigned long value to be converted to a string of decimal
  *   digits.
  * \returns a pointer to a string converted from \p value.  Except for zero,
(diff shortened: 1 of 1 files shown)


src/mmfatl.c
```diff
@@ -10,21 +10,49 @@
...
 
+#include <limits.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <stdarg.h>
 #include <string.h>
 
 #include "mmfatl.h"
 
+#if CHAR_BIT > 8
```
digama0, line 21:
```suggestion
#if CHAR_BIT != 8
```

wlammen:
@digama0 fully C99 compatible again.  Ready for review and merge.

wlammen:
Most of latest commits are cosmetic changes (style improvement, comment correction etc), although one glitch was corrected, too.  Still ready for review.

wlammen:
@digama0 More than a month passed without any reaction.  Is there any chance this will be committed?  Or at least for me to learn why it is rejected?  For clarity I opened another PR as a preview where all this heads towards.  As a remembrance:  The current code base is NOT in full compliance to what is expected from clean C11 code.  In particular, the error handling code relies on system code not eligible for use under harsh conditions like memory exhaustion.  In addition, not separating the error handler from the rest of the program (allowing access through an interface only) creates cyclic dependencies that are not only difficult to document, but are a problem on their own (a routine calls the error handler due to some resource problems, which internally crashes due to similar problems, again calling the error handler, and so on).

wlammen:
> Or normalizing the code style and updating the main logic to be more maintainable, use fewer global variables, etc.

This is what I have in my mind.  I did clean up other projects before, including some as part of my profession.  I generally start with the technical infrastructure, before moving up to more interesting software layers.  This turned out to be effective in the end.  And you need to set up general strategies (like the tests in this PR) first, before proceeding.

And the current stuff is more interesting as may appear on the outside, since I was bitten by the cyclic dependencies already, when I used doxygen to document the memory allocation code.  It was a nightmare to elaborate on the exact circumstances under which it is called.

PR #105, opened 2022-10-30
fatal error, implementation done

wlammen:
.. but for now not called from Metamath.

The implementation is completely hidden behind an interface.  So this sub-module is detached from the rest of Metamath, and both sides can be developed independently from each other, unless one part breaks the interface contract.

Note that this code is helpful, if it runs as desired, but it is not a critical part of Metamath.  It covers the fatal error path, which, if dysfunctional, is not really beyond annoying.  One could argue that the level of sophistication here is an overkill for this task.  Nevertheless, the self test suite and structured design technique would be nice if elsewhere employed as well.  During development this approach already caught bugs in an early phase, and my confidence in the code is high.  It successfully passed a compile with gcc flags -std=c99 and -pedantic.

What is still missing?  
1. It should be called from out-of-memory locations in Metamath.
2. Doxygen has not been tested with the comments.
3. UTF-8 support.

This will be part of the next pull request.
@digama0 Would be nice if at least one of the pull requests gets reviewed.  This pull request covers all of #104 .

### DIFF
--- a/src/mmfatl.c
+++ b/src/mmfatl.c
@@ -155,6 +155,14 @@ struct ParserState {
 
 static struct ParserState state;
 
+// reflect buffer overflow in the parser state
+static bool checkOverflow() {
+  bool result = !isBufferOverflow();
+  if (!result)
+    state.format += strlen(state.format);
+  return result;
+}
+
 /*!
  * initializes \ref state.
  * \post establish the invariant in state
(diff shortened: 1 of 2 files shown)


wlammen:
That's it, nothing more today.

wlammen:
Nothing more today

PR #110, opened 2022-12-08
fatal errors part 8

wlammen:
adds a handler for usual text (not a placeholder) in a format string

### DIFF
--- a/src/mmfatl.c
+++ b/src/mmfatl.c
@@ -142,23 +142,25 @@ static struct Buffer buffer;
  * We do not rely on any initialization during program start.  Instead we
  * assume the worst case, a corrupted pointer overwrote the buffer.  So we
  * initialize it again immediately before use.
- * \post \ref buffer is initialized 
+ * \post \ref buffer is initialized
+ * \param buffer [not null] the output buffer to empty and initialize
  */
-static void initBuffer(void) {
+static void initBuffer(struct Buffer* buffer) {
   char ellipsis[] = MMFATL_ELLIPSIS;
 
-  buffer.begin = buffer.text;
-  buffer.end = buffer.text + MMFATL_MAX_MSG_SIZE;
-  memset(buffer.begin, NUL, MMFATL_MAX_MSG_SIZE);
-  memcpy(buffer.end, ellipsis, sizeof(ellipsis));
+  buffer->begin = buffer->text;
+  buffer->end = buffer->text + MMFATL_MAX_MSG_SIZE;
+  memset(buffer->begin, NUL, MMFATL_MAX_MSG_SIZE);
+  memcpy(buffer->end, ellipsis, sizeof(ellipsis));
 }
 
 /*!
+ * \param buffer [const, not null] the buffer to check for overflow.
  * \return true, iff the current contents exceeds the capacity of the
  *   \ref buffer, so at least the terminating \ref NUL is cut off, maybe more.
  */
-inline static bool isBufferOverflow(void) {
-  return buffer.begin == buffer.end;
+inline static bool isBufferOverflow(struct Buffer const* buffer) {
+  return buffer->begin == buffer->end;
 }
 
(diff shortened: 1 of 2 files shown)


src/mmfatl.c
```diff
@@ -262,6 +261,19 @@ static void fatalErrorInit(void) {
...
 
+/*!
+ * copy text verbatim from a format string to the message buffer, until either
+ * the format ends, or a placeholder is encountered.
+ * \post member format of \ref state is advanced over the copied text, if no
+ *   overflow.
+ * \post member format of \ref state points to the terminating \ref NUL on
+ *   overflow.
+ */
+static void handleText(void) {
```
digama0, line 272: Why doesn't this have any arguments? This is the kind of C function I hate to see, relying on lots of global state.

digama0:
I would be very cautious about adding UTF-8 support in any form. The Metamath specification itself says that non-ASCII is not allowed, so it is not clear where you would be able to add UTF-8 support that doesn't just complicate matters when it comes to preventing its use in the appropriate places.

src/mmfatl.c
wlammen: One could add a pointer to the ParserState, but that is a global instance anyway.  Maybe one can create the instance on the stack, and pass it around, which I would usually do, but we are in a critical and memory tight situation, possibly including the stack, forcing us to use globals.  I'd love to throw an exception to get a clean stack again, but C 99 to my knowledge has no such mechanism.

I'll think it over, whether some minor unforced stack usage is acceptable.

wlammen:
Fortunately UTF-8 is an extension of ASCII, so a silent and partial transition won't hurt.  But you're right:  Some discipline in the transition period is called for.

digama0:
No, what I mean is, it is not to spec if you allow UTF-8 in e.g. comments. You are supposed to throw an error if you see any 8th bits set in the whole file.

wlammen:
The question is whether you want to leave it that way.  The strict ASCII limitations were from a time where a zoo of ASCII, Latin1, Latin2 character sets were in place concurrently.  Unicode has put an end to this more than a decade ago, and it offers a much better user experience.

If handling this is contentious, I won't extend code in this direction, even if it is possible at low cost.

digama0:
This is a much bigger question than just a metamath-exe. If you change the spec, you have to update all the verifiers, and I'm not sure either of us is ready to take that effort on. In particular, I think that it would be a move in the wrong direction with respect to metamath being a stupidly simple verification system that you can write in just a couple lines or implement on a turing machine if you want. Metamath is simple *on purpose*.

wlammen:
Let me answer with a vision here:  I fancy Metamath be a tool, a library that you can use like git, with a plethora of selectable and independent primitive tasks.  Instead, it presents us with a separate workspace, that is not built on top of the mentioned library.  If we could go in that direction, I envision a far more flexible future than we have now.

Each of these library primitives could offer UTF-8 on an individual base.  You still may introduce a strict mode, where such flexibility is disallowed.

src/mmfatl.c
wlammen: I agree with you that generally the signature of a function should expose
- which objects are read
- which are modified.
This is for better readability and maintenance.
In our case objects are limited to two, namely a Buffer and a ParserState.  For the sake of a good coding style I 'localized' the State, but did not do so with Buffer, since this is clearly a singleton.
The State is not likely to appear in many instances, too, or be created on the stack, since it is heavyweight as a container for the values replacing placeholders, but this may change.  The values are a constant block of data, not subject to change, only the current position is important.  So fine-tuning may remove the values from the State in future.

wlammen:
@digama0 adapted the signatures of functions such that the dependencies are exposed.

wlammen:
@digama0 It would be nice to get this merged, so I can use the extra spare time over Christmas to continue (and hopefully finish it before 2023).

PR #114, opened 2022-12-30
remove VAX-related code (and rewrap some overlong lines)

benjub:
Remove VAX-related code, which is a task from #17.
(I did global searches on "vax" and "vms", and also on "maci", "mmac", "think" to check no code related to Think C or the old mmmaci.c remained, except possibly in comments.)
Also rewrap some overlong lines.


benjub:
@digama0 : si I think you can check the box "Delete VAXC code (VAX machines are long out of use)" in #17: 4 done, 32 to go !

PR #117, opened 2023-01-06
refactor #1 explain the design principles of mmfatl

wlammen:
In the past days I ramped up my knowledge about refactoring again. Refactoring is done in wlammen/metamath-exe, branch wl, which is currently a copy of master. From there I will regularly transfer commits to metamath-exe/master, but I won't wait for any acceptance any more. If PRs get delayed, well, then the tip of my branch will slowly grow into a genuine fork.

Refactoring means you improve the code design and functionality bit by bit. Each commit addresses only one isolated aspect of the software, and the preferred small changes will mostly be such, that the code transition will keep the code semantics intact. This will make review fairly easy.

The basic ideas behind my refactoring are notified in the file refactor.log, that hopefully allow people to grasp the motivation and implementation of ideas from a bird eye's perspective.

### DIFF
--- a/refactor.log
+++ b/refactor.log
@@ -1,0 +1,8 @@
+======================   Refactor log   ======================
+
+Protocol of the changes applied to the executable 'metamath'.  Each change
+should correspond to a small set of commits implementing this change, and
+nothing else (apart from cosmetic corrections).
+
+[wl c6a76d9] Comment update: Better explain the design principles of the fatal
+   error handler in the 'Rationale' section.
--- a/refactor.log
+++ b/refactor.log
@@ -1,8 +1,0 @@
-======================   Refactor log   ======================
-
-Protocol of the changes applied to the executable 'metamath'.  Each change
-should correspond to a small set of commits implementing this change, and
-nothing else (apart from cosmetic corrections).
-
-[wl c6a76d9] Comment update: Better explain the design principles of the fatal
-   error handler in the 'Rationale' section.
--- a/src/mmfatl.h
+++ b/src/mmfatl.h
@@ -41,10 +41,16 @@
  * message.
  *
  * To display this final message, we restrict its code to very basic,
- * self-contained routines, independent of the rest of the program to the
- * extent possible, thus avoiding any corrupted data.
+ * self-contained routines.  It does not share state with the rest of the
+ * program, and uses shared resources only to the bare minimum:
+ * (a) needs a few hundred bytes of stack memory, i.e. the most basic execution
+ *     environment capable of calling a library function;
+ * (b) expects only a simple library function forwarding 1 KB of raw text to
+ *     stderr being operational.
+ * In particular memory corruption in the Metamath executable cannot affect the
+ * handling of the fatal error.
  *
- * In particular everything should be pre-allocated, so the risk of a failure
+ * To achieve this everything should be pre-allocated, so the risk of a failure
  * in a corrupted or memory-tight environment is minimized.  This is to the
  * detriment of flexibility, in particular, support for dynamic behavior is
  * limited.  Many Standard C library functions like \p printf MUST NOT be
@@ -68,8 +74,8 @@
  * In a memory tight situation we cannot reset the memory heap, or stack, to
  * have free space again for, say, \p printf, even though we are about to exit
  * program execution, for two reasons:
- *   - We want to gather diagnostic information, so the program structures need
- *     to be intact;
+ *   - We want to gather diagnostic information, so some program structures
+ *     need to be intact;
  *   - The fatal error routines need not be the last portion of the program
  *     executing.  If a function is registered with \p atexit, it is called
  *     after an exit is triggered, and this function may rely on allocated


refactor.log
```diff
@@ -0,0 +1,8 @@
+======================   Refactor log   ======================
+
+Protocol of the changes applied to the executable 'metamath'.  Each change
+should correspond to a small set of commits implementing this change, and
+nothing else (apart from cosmetic corrections).
+
+[wl c6a76d9] Comment update: Better explain the design principles of the fatal
+   error handler in the 'Rationale' section.
```
digama0, line 8: Why are you introducing yet another way to track changes? There is already a changelog in metamath.c.

wlammen: This is temporary, focusing on my work, and more fine grained.  Since refactoring is about small steps, the change log would grow extremely, even protocolling things that get reversed shortly after.  Do you want to see an entry for this commit for example? If you wish, I will remove it again, but that will take effect only after the currently pending commits are done.

digama0: I would prefer you put those kind of fine-grained comments in the commit messages themselves, or the PR description. The code itself should not have temporary stuff in it, leave that in your own local copy.

wlammen:
I don't know what to do.  When I submit an all-in-one contribution like #86 , it is ignored for months.  If I break it up into easy-to-review smaller steps, I get bashed for continuously adding dead code, although this is only temporary.  I modified my working process again by doing development first in my private, but publicly visible repository, so you can preview the upcoming steps while each individual step like this one still remains simple, allowing for fast review and the necessary detail discussion.  This resembles the process in our company to some extent, where we have releases every other week, one that gets special attention and when the change-log is updated.  On the other hand I won't roll back all my private work just to get a typo fixed.  Such cosmetic changes can be done at the current tip of my repository, from where it will eventually find its way into the official line.  How do you think we should proceed without putting an unwieldy administrative burden on me?

digama0:
> I modified my working process again by doing development first in my private, but publicly visible repository, so you can preview the upcoming steps while each individual step like this one still remains simple, allowing for fast review and the necessary detail discussion.

This sounds great, although just FYI I probably won't be looking at that repo unless you specifically direct me to it.

> On the other hand I won't roll back all my private work just to get a typo fixed. Such cosmetic changes can be done at the current tip of my repository, from where it will eventually find its way into the official line. How do you think we should proceed without putting an unwieldy administrative burden on me?

What I do in similar cases is to have a local branch where I have a bunch of wide sweeping changes, more than I expect will make it past PR in one go or maybe ever, and then after making sure that it mostly works I take off a piece of that, for example all of the refactors that touch one function or one file, and PR that. The PR is completely unrelated to my local branch, so no one has to see my dirty laundry and I certainly don't make any ultimatum along the lines that the dirty laundry has to be merged for the refactor to proceed.

When the PR is merged, I rebase my branch on top of the new master; most likely this only has the effect of removing some of my changes from the branch and leaving my notes and such in place. Rinse and repeat.

digama0:
As regards this PR specifically, it is an easy merge if you remove the refactor log.

wlammen:
You won't see much dirty laundry in my publicly visible repository, that's already done on the side.
It was a considerate decision to add refactor.log, because it allows to track the fundamental ideas across several PRs, allowing each one be simple (frog's view) without losing contact to the bird's view.  Whether this is a workable concept - well, there could at least have been some evaluation period.  In the end, refactoring is an incremental process, where some intermediate steps may result in a local degradation, for the sake of the bigger picture later.
I am not a fan of having a plethora of small sized commits, from which I (cherry-)pick dozens for a PR, hopefully without missing one important.  That said, I will introduce another stage on my computer between the laundry level and the one from which generate commits to my private repository, from which I design my personal origin, being the base for PRs to metamath-exe.  Let's see, how this pans out.

digama0:
> It was a considerate decision to add refactor.log, because it allows to track the fundamental ideas across several PRs, allowing each one be simple (frog's view) without losing contact to the bird's view. Whether this is a workable concept - well, there could at least have been some evaluation period.

You should keep separate PRs that add a new thing from PRs that modify an existing thing, since they have very different standards for evaluation and require different justification in the PR description and subsequent discussion. If you want to add a refactor log, do it in it's own PR and explain why this is a good idea for us to use, what we should put in it, and how to keep it correct.

I'm assuming that *this* PR is about the docs modification, which looks uncontentious. Mixing it with an inadequately justified new file results in a much harder review.

wlammen:
> I'm assuming that this PR is about the docs modification, ...

It was more the other way round: I am much more concerned with establishing a functioning working process, and made suggestions how to improve, or at least test new ways.  I could not have added refactor.log without any contents at all, so for demonstration purposes I looked for something uncontentious to fill with. See all my notes above, where I do not mention the doc change at all.
Again we have a communication problem here.
As for the orthogonality I am on your side, but would like to exempt cosmetic changes to reduce administrative load.

digama0:
I see. In that case my earlier comment is appropriate: the PR description does not explain why we need a file in the repo for this. Stuff in the repo is not just for your personal use, so we need to have a process for making everyone use it correctly and I don't see how that could make anything easier. We have git / github already for making logging-like descriptions of changes; I don't see any point for a changelog / "refactor log" except to tell users about new features and behavior changes (which is the classical purpose of a changelog). A refactor log doesn't sound like it would be important for users, so I'd rather not have it at all.

wlammen:
My problem is that I feel as if I am back on square one again.  The same situation as at the beginning of the mmfatl series.  Well not exactly.  I am now bitten by a nasty experience, which earned me no appreciation, just trouble.  If I continued contributing in the same way, why wouldn't I end up in more or less the same mess again?  I need something more promising here.

digama0:
A simple way to avoid the issue, which may or may not be acceptable depending on your constraints or desires, is to not PR anything that is without precedent in the codebase. New things are hard because they expand the gamut of what is required to understand the code as a whole. Modifications to existing things are much easier, like the doc modification itself in this PR.

Doxygen was a new thing, so there was some work to add it, but now it's not new anymore so you can add doxygen comments to any function and this will not raise any eyebrows. (The comments should still be correct, of course.)

Refactors in the sense of moving functions around, fixing indentation and so on are unlikely to fall afoul of this guideline. (You still can, by adding new and interesting ways to capitalize types or using tabs vs spaces, but as long as you do what is done somewhere else in the code you should be fine.)

Bug fixes are almost certainly fine, in the sense of small deltas on the code that solve issues. `mmfatl` can be conceptualized as a bugfix, but it fell afoul of this rule because it was a new module, a new methodology on code comments, and arguably a new approach to evaluation of code correctness in the rest of the project, without precedent in the codebase.

I really think this is not a high bar. It is a conservative rule, but I think that is more useful for you if you are trying to predict how much trouble a PR will be to merge.

wlammen:
Metamath's code base was "afoul" of external rules written down in the C standard at a very fundamental level.  This is imo difficult to fix without partially leaving your "comfort zone" (i.e. getting away without any new method).  Don't forget that part of the discussion took place in the precursor #86 already, and that the test suite was added months after the initial code was published without avail.  But I agree, that the issues section could be the better platform in future.

I still miss a word about intermediate results.  Look at a refactoring like the one in https://github.com/metamath/metamath-exe/commit/aa2b9e508d25b355b8f934d441763e851e24be2d where (1) outOfMemory is moved to mmfatl, (2) it is covered by the test suite, (3) it's signature is adapted to take advantage of the built-in features of the error handler, so callers do not create cyclic dependencies by calling cat().  I don't recommend a single PR here, for the amount of changes, that could spark another nasty discussion.  But then you have to live with intermediate results, that do nothing useful at the moment of publication.  What will be your stance then?

digama0:
> Metamath's code base was "afoul" of external rules written down in the C standard at a very fundamental level. This is imo difficult to fix without partially leaving your "comfort zone" (i.e. getting away without any new method).

This is true. Change is hard, but sometimes worthwhile. Not changing is easier though, and you should be aware of that when planning how much to spend on discussion vs code changes.

> I still miss a word about intermediate results. Look at a refactoring like the one in https://github.com/metamath/metamath-exe/commit/aa2b9e508d25b355b8f934d441763e851e24be2d where (1) outOfMemory is moved to mmfatl, (2) it is covered by the test suite, (3) it's signature is adapted to take advantage of the built-in features of the error handler, so callers do not create cyclic dependencies by calling cat(). I don't recommend a single PR here, for the amount of changes, that could spark another nasty discussion. But then you have to live with intermediate results, that do nothing useful at the moment of publication. What will be your stance then?

I don't see any issues with that PR. It passes the "no new things" test (except for the refactor.log change since refactor.log was removed from this PR) because mmfatl is not new anymore.

PR #118, opened 2023-01-08
consistent comment after #endif directives (in files mmfatl.*)

benjub:
I tried to make comments after `#endif` more consistent by systematically recalling the identifier that was after the mathching `#if(n)def` directive.  Please @wlammen tell me if this is correct and if I can do that for the other files.

What about removing double blank lines everywhere ? (regex `\n\n\n`)

### DIFF
--- a/src/mmfatl.c
+++ b/src/mmfatl.c
@@ -439,10 +439,10 @@ void fatalErrorPrintAndExit(void) {
       && !isBufferEmpty(buffer)
       && getLastBufferedChar(buffer) != LF)
     fatalErrorPush("\n");
-#ifndef TEST_ENABLE // we don't want a test program terminate here
+#ifndef TEST_ENABLE // we do not want a test program terminating here
   fputs(buffer->text, stderr);
   exit(EXIT_FAILURE);
-#endif
+#endif // TEST_ENABLE
 }
 
 void fatalErrorExitAt(char const* file, unsigned line,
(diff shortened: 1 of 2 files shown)


wlammen:
It is grammatically correct, and can be used to increase readability when the framing #if and #else #endif are more than a few lines apart, or are heavily stacked, so you might get the corresponding pair wrong.  Maybe it is used for automatic code maintenance using simple scripts, too. Otherwise the extra comments have a bit of a littering effect, imo.  So the particular examples here are questionable.  Would they have helped you grasp the code logic easier?

benjub:
I think everything that tends to more consistency makes it easier. I am new to the C language and I tried to have a look, but I'm daunted by the size of the code base...
Is it possible to run some kind of autoformatting tool to ensure consistency of indentations, linebreaks, etc. ?

wlammen:
Some C/C++ IDE like eclipse https://www.eclipse.org/downloads/packages/release/2022-12/r/eclipse-ide-cc-developers come with some kind of formatting support.

digama0:
> I think everything that tends to more consistency makes it easier.

Strong agree with this. We should strive for consistency in the codebase, even if the style is a bit weird - nothing is worse than a codebase with inconsistent *and* weird style. (If the style is bad, I also welcome changes to improve it, provided they are applied uniformly.)

digama0:
> What about removing double blank lines everywhere ? (regex `\n\n\n`)

Sounds good to me.

PR #124, opened 2023-01-15
Misc

benjub:
I'm not sure what this is: there were four "form feed" characters, each on its own line. They do not appear on github, so the diff looks empty.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -120,12 +120,12 @@ https://guides.lib.berkeley.edu/how-to-write-good-documentation
     an entry point to documentation a README.TXT is kept in the main folder of
     Metamath.  This file is pure [ASCII](https://en.wikipedia.org/wiki/ASCII)
     encoded [plain text](https://en.wikipedia.org/wiki/Plain_text).
-    
+
 The following rules are strong recommendations.  We encourage to follow them as
 much as possible.  But in the end we think that some sort of documentation is
 better than no documentation, so bypassing any of them is possible if somehow
 inevitable.
-    
+
 2. If your documentation is simple enough to be expressed as
     [plain text](https://en.wikipedia.org/wiki/Plain_text), we recommend to
     rely as much as possible on [ASCII](https://en.wikipedia.org/wiki/ASCII)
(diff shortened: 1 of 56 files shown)


digama0:
That sounds like `\r\n` newlines, which are the standard on windows. Git normalizes line endings to `\n` inside the repo and converts back and forth transparently when you use it on windows, but if you check in a file with `\r\n` newlines while the config flag for this behavior is disabled you can get `\r\n` in the repo files, which results in weird results on non-windows systems.

benjub:
Yes, it's form feeds. Displayed as `^L` on my terminal and something like $F_F$ (the unicode version) in VS Code (see https://en.wikipedia.org/wiki/ASCII#Control_code_chart).

I unexpectedly found those because it messes up some regex searches.

* Merge or close this PR ? (there are no other `\f` in the whole repo, and no `\r` nor `\v` at all)

* In the whole repo, the only files using tabulations `\t` are LICENSE.TXT and two Makefile.am (apart from files `.git*`). Change them to spaces or leave them as is ?

digama0:
Tabs in makefiles are a part of the syntax, you can't change those. For LICENSE.TXT I think it's fine as long as we don't change the text, so if you want to remove the `\f` we can do that. On the other hand it might make it easier for automated tools to recognize the license if we don't mess with it; github seems to be fine with it so far.

benjub:
@digama0 : I copied the license verbatim from https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt
(I do not know if it is worth using https://www.gnu.org/licenses/gpl-3.0.txt instead)

digama0:
GPLv2 vs v3 is a big debate in certain circles, so I would not make any modifications without more careful consideration.

jkingdon:
> The only clue I could find (from https://www.asciihex.com/character/control/12/0x0C/ff-form-feed) is that it might be used by some printers for page breaking and is used conventionally in some plain text printable files, like this license.

That's about right. People would use a form feed to delineate parts of a file (sort of like a section heading or something).

Even in the day I think this was somewhat rare, but today there is no reason to preserve it.

PR #128, opened 2023-01-24
shrink live ranges:  typeProof, cleanWrkProof

digama0:
I'm going to try to make a series of PRs like this which moves allocation/initialization of variables to just before first use and deallocation to just after last use, toward #17 "Move variable declarations from top of function to first use".

### DIFF
--- a/src/mmcmds.c
+++ b/src/mmcmds.c
@@ -1339,46 +1339,6 @@ void typeProof(long statemNum,
         - determined by the global variable g_midiFlag, not by a parameter to
         typeProof()
   */
-  long i, j, plen, step, stmt, lens, lent, maxStepNum;
-  vstring_def(tmpStr);
-  vstring_def(tmpStr1);
-  vstring_def(locLabDecl);
-  vstring_def(tgtLabel);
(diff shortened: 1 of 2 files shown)


jkingdon (approved):
This strikes me as relatively safe (well as much as you can say that about code which is about memory allocation) and I did go over it moderately carefully. The biggest caveat is just that it touches a lot of variables so I don't feel super well calibrated about exactly how cautious to be.

david-a-wheeler:
I haven't reviewed the details, but the general principle sounds great. Historically C only allowed variable declaration at the beginning of a scope (open curly brace), which encouraged "early" allocation.

It might be helpful to compile this with address sanitizer (`-fsanitize=address`) and run a bunch of tests. (I recommend enabling `-O2` for performance and `-fno-omit-frame-pointer` so the error reports are understandable). That is much slower & takes more memory, so you don't want to do it in "production", but it can detect a lot of problems.

digama0:
That sounds like a good idea for the CI. The whole process is still less than a minute so we can definitely afford to run it with more instrumentation.

PR #130, opened 2023-02-20
tex generation: avoid '\rm', load mathtools and phonetic

benjub:
* tex generation: avoid '\rm', load mathtools and phonetic
* comment formating

There remain some `\tt` that should be replaced by either `texttt{` or `\mathtt{` depending on whether we are in text or math mode (was not obvious at first in the C code).

Also, I added the `phonetic` package, so that we can replace the latexdef's using \rotate with \riota, and then one can remove the `graphicx` package.

### DIFF
--- a/src/mmdata.h
+++ b/src/mmdata.h
@@ -946,4 +946,4 @@ int qsortStringCmp(const void *p1, const void *p2);
 /*! Call on exit to free memory */
 void freeData(void);
 
-#endif /* METAMATH_MMDATA_H_ */
+#endif // METAMATH_MMDATA_H_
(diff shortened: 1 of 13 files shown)


benjub:
Is it good for you @digama0 ?

benjub:
Looks good to you @digama0 ?

The package `mathrsfs` that I add in this PR is necessary for the LaTeX files generated by metamath.c to compile. I added the package `mathtools` for better spacing and the package `phonetic` to be able to typeset `riota` more simply. Finally, I removed the occurrences of the deprecated `\rm`. I did not remove the deprecated `\tt` since I could not be sure they appear in math mode (in which case they should be `\mathtt`) or text mode (hence `\texttt`).

digama0:
Can you move the style changes to a separate PR?

benjub:
> Can you move the style changes to a separate PR?

Done in #131.

PR #131, opened 2023-03-05
tex generation: \rm-->\mathrm and \usepackage{...}

benjub:
This is the non-style part of #130.


digama0:
It would be nice if we had a test case for tex generation (not checking that it compiles, but just a straight input-output test) so we can make sure it works. But making the test suite handle tex tests is more work than I think should go in this PR so I'll merge it as is.

PR #138, opened 2023-05-10
Line comments instead of block comments

GinoGiotto:
In https://github.com/metamath/metamath-exe/issues/17 I read:
> - [ ]  Use // comments instead of /* (or doc-comments where appropriate)

It doesn't seem a difficult task, so I thought I could give my contribution.

I might continue this if I did a good change.

### DIFF
--- a/src/mmword.c
+++ b/src/mmword.c
@@ -105,14 +105,13 @@ void revise(FILE *f1_fp, FILE *f2_fp, FILE *f3_fp, vstring addTag, long m)
   let(&l2_,ctlz_);
   eof1=eof2=0; // End-of-file flags
   d=0;
-
-l7100:  // Lines 7100 through 7300 are a modified version of the compare loop
-        // in DEC's FILCOM.BAS program.
-
+// Lines 7100 through 7300 are a modified version of the compare loop in DEC's
+// FILCOM.BAS program.
+l7100:  
         if (!strcmpe(l1_,l2_)) { // The lines are the same
                 if (strcmpe(l2_,ctlz_)) {
-                  fprintf(f3_fp_, "%s\n", l2_); // Use edited version
-                                                // of line when they are the same.
+                  // Use edited version of line when they are the same.
+                  fprintf(f3_fp_, "%s\n", l2_);
                 }
                 gosub_7320();
                 gosub_7330();
@@ -145,8 +144,8 @@ l7100:  // Lines 7100 through 7300 are a modified version of the compare loop
                         return;
                 }
         }
-        d=d+1; // Number of difference sections found so far
-               // (For user information only)
+        // Number of difference sections found so far (For user information only)
+        d=d+1; 
         i1_=i2_=m-1;
         let(&line1_[0],l1_);
         let(&line2_[0],l2_);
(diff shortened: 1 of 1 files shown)


GinoGiotto:
I'm doing a mix of python and manual changes (like adding periods in some circumstances to clarify the end of a sentence). Sometimes there are odd intentations and right now I'm just using my personal taste.

If there are strict aesthetic rules, let me know bc they would help me automatize the process even more.

src/mmword.c
```diff
@@ -288,8 +286,8 @@ l7220:  i=i+m-1;
              fprintf(f3_fp_, "%s\n", tmpLine);
            }
            fprintf(f3_fp_, "%s\n", line1_[i]);
-                                     /* Output original deleted lines */
-           /*freeTempAlloc();*/ /* Clear vstring stack */
+                                     // Output original deleted lines
```
digama0, line 289: BTW, the reason for the funny indenting here IIUC is that they are *right aligned*, to 79-80 characters. I think we can just stop doing this, I much prefer putting comments about a line of code on the previous line if it doesn't fit in the margin.

GinoGiotto: I made another commit, is this what you mean?

src/mmword.c
```diff
@@ -285,9 +284,9 @@ l7220:  i=i+m-1;
                  addTag_, 0);
              fprintf(f3_fp_, "%s\n", tmpLine);
            }
+           // Output original deleted lines // freeTempAlloc();
+           // Clear vstring stack
```
digama0, line 288: This comment got mangled, the freeTempAlloc() should go after the printf, and the "clear vstring" comment applies to it

GinoGiotto: Done

src/mmword.c
```diff
@@ -341,70 +338,70 @@ l7220:  i=i+m-1;
 }
 
 void gosub_7320(void) {
-        /* Subroutine:  get next L1_ from original file */
+  // Subroutine:  get next L1_ from original file
```
digama0, line 341: This comment applies to the previous line, and should go before it. Dunno if you are adventurous enough to try automatically detecting it, but a fairly good heuristic would be that the start of the comment is more indented than the following line.

src/mmword.c
```diff
@@ -341,70 +338,70 @@ l7220:  i=i+m-1;
...
-  let(&l1_, cat(tmpLin, l1_, NULL)); /* Add any blank lines */
-  free_vstring(tmpLin); /* Deallocate */
+  let(&l1_, cat(tmpLin, l1_, NULL)); // Add any blank lines
+  free_vstring(tmpLin); // Deallocate
   return;
 }
 
 void gosub_7330(void) {
-        /*  Subroutine:  get next L2_ from edited file */
+  // Subroutine: get next L2_ from edited file
```
digama0, line 375: ditto

digama0 (approved):

src/mmword.c
GinoGiotto: Right now I'm leaning more for changing those manually, but I'll give a look and see if they are common enough to be worth automating.

PR #139, opened 2023-05-13
Line comments instead of block comments 2

GinoGiotto:
Continuation of #138.

As I was viewing the code, I noticed certain types of comments that I may not be supposed to edit, so I chose to omit them. These comments can be categorized into four groups:

1. Comments with `/*!***/`: They seem to have a special type of formatting, which is consistent with some C files I came across that were not created by Norm. Specifically, I looked at `mmfatl.c` by @wlammen and `mmtest.c` by @digama0, both of which have comments already updated in the `//` form, except for the `/*!***/` ones, so they seem to be intentionally built this way.

2. Comments that have multiple `***`: It appears that they are intended to be noticeable, for example like the header that is present at the beginning of every file:
<pre>/*****************************************************************************/
/*        Copyright (C) 2018  NORMAN MEGILL                                  */
/*            License terms:  GNU General Public License                     */
/*****************************************************************************/
/*34567890123456 (79-character line to adjust editor window) 2345678901234567*/</pre>
Or like the following comment:
<pre> /*** See the comments in mmvstr.h for an explanation of these functions ******/ </pre>
So I left those out too (the header has the same formatting in the already updated `mmfatl.c` `mmtest.c` files too).

3. Comments that are inside the code: These are probably the most uncontroversial ones since they wouldn't work with `//`, unless with some rewrapping of the code itself.
Example:
<pre> if (graphicsChar == 218 /* up left */ || graphicsChar == 217 /* lo r */
          || graphicsChar == 191 /* up r */ || graphicsChar == 192 /* lo l */)</pre>
Technically the comment `/* lo r */` isn't strictly between coding characters, so it could be edited with `// lo r`, but it's clearly of the same type of the neighbouring ones, so I left it out too.

4. Multiline commented out C code: It appears that it was intended to be easily uncommented for experimental use. Since line comments would be less practical for this purpose, I left the original block ones.
Example:
<pre>// For debugging
/*

int main(void)
{
  vstring_def(s);
  vstring_def(t);

  printf("Hello\n");
  let(&t,edit(" x y z ",2));
  let(&s,cat(right("abc",2),left("def",len(right("xxx",2))),"ghi",t,NULL));
  printf("%s\n",s);
  printf("num %s\n",num(5));
  printf("str %s\n",str(5.02));
  printf("num1 %s\n",num1(5.02));
  printf("time_ %s\n",time_());
  printf("date %s\n",date());
  printf("val %f\n",val("6.77"));
}

*/</pre>
In general my convention knowledge for C is very limited, so I built these criteria by intuition. Let me know if you agree with these criteria or if there are some standards that you are aware of.

### DIFF
--- a/src/mmvstr.c
+++ b/src/mmvstr.c
@@ -16,18 +16,17 @@ This is an emulation of the string functions available in VMS BASIC.
 #include <string.h>
 #include <time.h>
 #include "mmvstr.h"
-/*E*/ /*Next line is need to declare "db" for debugging*/
+/*E*/ // Next line is need to declare "db" for debugging
 #include "mmdata.h"
-/* mmdata.h is also used to declare the bug() function that is called in
-   several places by mmvstr.c.  To make mmvstr.c and mmvstr.h completely
-   independent of the other programs, for use with another project, do the
-   following:
-     (1) Remove all lines beginning with the "/ *E* /" comment.
-     (2) Remove all calls to the bug() function (4 places).
-   To see an example of stand-alone usage of the mmvstr.c functions, see
-   the program lattice.c and several others included in
-     http://us.metamath.org/downloads/quantum-logic.tar.gz
-*/
+// mmdata.h is also used to declare the bug() function that is called in
+// several places by mmvstr.c.  To make mmvstr.c and mmvstr.h completely
+// independent of the other programs, for use with another project, do the
+// following:
+//   (1) Remove all lines beginning with the "/ *E* /" comment.
+//   (2) Remove all calls to the bug() function (4 places).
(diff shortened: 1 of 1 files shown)


digama0:
> 1. Comments with `/*!***/`: They seem to have a special type of formatting, which is consistent with some C files I came across that were not created by Norm. Specifically, I looked at `mmfatl.c` by @wlammen and `mmtest.c` by @digama0, both of which have comments already updated in the `//` form, except for the `/*!***/` ones, so they seem to be intentionally built this way.

These are doxygen comments. I think doxygen supports a variety of styles but if you want to be sure you should test using `./build.sh -d`.

PR #140, opened 2023-05-13
Simplify and harden `switchPos`

### LINKED ISSUE #136: Minimize bugs
While I was experimenting with the minimize command I had the idea to allow introduction of `|-` definitions and `wff` definitions. I noticed that `wff` definitions in set.mm start with `w*`, so I tried the command:
<pre>MM-PA> MINIMIZE_WITH * /ALLOW_NEW_AXIOMS df-* w*
                                              ^^
?Expected / or nothing.
MM-PA></pre>

This produces an error, so it seems to be  an invalid request.

However if I write the same thing in this form:
<pre>MM-PA> MINIMIZE_WITH * /
VERBOSE, MAY_GROW, EXCEPT, OVERRIDE, INCLUDE_MATHBOXES, ALLOW_NEW_AXIOMS,
 NO_NEW_AXIOMS_FROM, FORBID, or TIME <VERBOSE>? ALLOW_NEW_AXIOMS
What statement label match pattern? df-* w*
/ or nothing &lt;nothing&gt;?
Bytes refer to compressed proof size, steps to uncompressed length.
Scanning forward through statements...
No shorter proof was found.
MM-PA>
</pre>

Then it doesn't complain and starts scanning, which is inconsistent with the behaviour of the first version of the command.

Not only this behaviour is inconsistent, but it also generates bugs in some circumstances, I was able to find 2 of them, the first one:
<pre>MM-PA> MINIMIZE_WITH * /
VERBOSE, MAY_GROW, EXCEPT, OVERRIDE, INCLUDE_MATHBOXES, ALLOW_NEW_AXIOMS,
 NO_NEW_AXIOMS_FROM, FORBID, or TIME <VERBOSE>? ALLOW_NEW_AXIOMS
What statement label match pattern? df-* w*
/ or nothing &lt;nothing&gt;? /
VERBOSE, MAY_GROW, EXCEPT, OVERRIDE, INCLUDE_MATHBOXES, ALLOW_NEW_AXIOMS,
 NO_NEW_AXIOMS_FROM, FORBID, or TIME <VERBOSE>? FORBID
What statement label match pattern? ax-13
/ or nothing &lt;nothing&gt;?
</pre>

Which generates the following msgbox. Pressing the OK button will make the program close by itself. In this case Metamath doesn't start minimizing and it doesn't even produce a bug check.

forbid

The second circumstance that generates this bug is:
<pre>MM-PA> MINIMIZE_WITH * /
VERBOSE, MAY_GROW, EXCEPT, OVERRIDE, INCLUDE_MATHBOXES, ALLOW_NEW_AXIOMS,
 NO_NEW_AXIOMS_FROM, FORBID, or TIME <VERBOSE>? ALLOW_NEW_AXIOMS
What statement label match pattern? df-* w*
/ or nothing &lt;nothing&gt;? /
VERBOSE, MAY_GROW, EXCEPT, OVERRIDE, INCLUDE_MATHBOXES, ALLOW_NEW_AXIOMS,
 NO_NEW_AXIOMS_FROM, FORBID, or TIME <VERBOSE>? EXCEPT
What statement label match pattern? ax-13
/ or nothing &lt;nothing&gt;? 
</pre>
which produces this msgbox (and then Metamath closes by itself when the OK button is pressed, as in the first example).

except

These bugs presents themselves in this form in the 0.198 version. In the 0.199 one the program just crashes without showing any msgbox.

As a side note I guess the right way to allow both `|-` definitions and `wff` definitions would be:
<pre>MM-PA> MINIMIZE_WITH * /ALLOW_NEW_AXIOMS df-* /ALLOW_NEW_AXIOMS w*</pre>

but I'm not sure if this kind of request was contemplated in the design. Does metamath takes care of both  `/ALLOW_NEW_AXIOMS` or one nullifies the other? (question for Metamath developers).

Finally here is an example of a complete log sequence to reproduce one version of these issues:
<pre>Metamath - Version 0.199.pre 29-Jan-2022      Type HELP for help, EXIT to exit.
MM> READ set.mm
Reading source file "set.mm"... 43107244 bytes
43107244 bytes were read into the source buffer.
The source has 204913 statements; 2734 are $a and 40822 are $p.
No errors were found.  However, proofs were not checked.  Type VERIFY PROOF *
if you want to check them.
MM> PROVE bezout
Entering the Proof Assistant.  HELP PROOF_ASSISTANT for help, EXIT to exit.
You will be working on statement (from "SHOW STATEMENT bezout"):
$d x y A $.  $d x B y $.
51216 bezout $p |- ( ( A e. ZZ /\ B e. ZZ ) -> E. x e. ZZ E. y e. ZZ ( A gcd B
      ) = ( ( A x. x ) + ( B x. y ) ) ) $= ... $.
Note:  The proof you are starting with is already complete.
MM-PA> MINIMIZE_WITH * /
VERBOSE, MAY_GROW, EXCEPT, OVERRIDE, INCLUDE_MATHBOXES, ALLOW_NEW_AXIOMS,
 NO_NEW_AXIOMS_FROM, FORBID, or TIME <VERBOSE>? ALLOW_NEW_AXIOMS
What statement label match pattern? df-* w*
/ or nothing &lt;nothing&gt;? /
VERBOSE, MAY_GROW, EXCEPT, OVERRIDE, INCLUDE_MATHBOXES, ALLOW_NEW_AXIOMS,
 NO_NEW_AXIOMS_FROM, FORBID, or TIME <VERBOSE>? FORBID
What statement label match pattern? ax-13
/ or nothing &lt;nothing&gt;?</pre>

However my assumption is that these behaviours will show no matter what database you use or what theorem you are trying to minimize.

digama0:
The original implementation concatenated all the arguments with spaces, which is not correct (causes a segfault) when the arguments themselves can have spaces. Here we use a much simpler mechanism, just searching for `/` and then `MY_SWITCH`, taking advantage of the fact that every existing switch is one word.

fixes #136

### DIFF
--- a/src/metamath.c
+++ b/src/metamath.c
@@ -1222,7 +1222,7 @@ void command(int argc, char *argv[]) {
           print2(
               "Warning:  You have not saved changes to the proof of \"%s\".\n",
               g_Statement[g_proveStatement].labelName);
-          if (switchPos("/ FORCE") == 0) {
+          if (switchPos("FORCE") == 0) {
             str1 = cmdInput1("Do you want to EXIT anyway (Y, N) <N>? ");
             if (str1[0] != 'y' && str1[0] != 'Y') {
               print2("Use SAVE NEW_PROOF to save the proof.\n");
@@ -1248,7 +1248,7 @@ void command(int argc, char *argv[]) {
       } else {
         if (g_sourceChanged) {
           print2("Warning:  You have not saved changes to the source.\n");
-          if (switchPos("/ FORCE") == 0) {
+          if (switchPos("FORCE") == 0) {
             str1 = cmdInput1("Do you want to EXIT anyway (Y, N) <N>? ");
             if (str1[0] != 'y' && str1[0] != 'Y') {
               print2("Use WRITE SOURCE to save the changes.\n");
@@ -1304,7 +1304,7 @@ void command(int argc, char *argv[]) {
       let(&g_commandFileName[g_commandFileNestingLevel], g_fullArg[1]);
 
       g_commandFileSilent[g_commandFileNestingLevel] = 0;
-      if (switchPos("/ SILENT")
+      if (switchPos("SILENT")
           || g_commandFileSilentFlag /* Propagate silence from outer level */) {
         g_commandFileSilent[g_commandFileNestingLevel] = 1;
       } else {
@@ -2070,7 +2070,7 @@ void command(int argc, char *argv[]) {
 
       readInput();
 
-      if (switchPos("/ VERIFY")) {
+      if (switchPos("VERIFY")) {
         verifyProofs("*",1); /* Parse and verify */
       } else {
         /* verifyProofs("*",0); */ /* Parse only (for gross error checking) */
@@ -2079,7 +2079,7 @@ void command(int argc, char *argv[]) {
       if (g_sourceHasBeenRead == 1) {
         if (!g_errorCount) {
           let(&str1, "No errors were found.");
-          if (!switchPos("/ VERIFY")) {
+          if (!switchPos("VERIFY")) {
               let(&str1, cat(str1,
          "  However, proofs were not checked.  Type VERIFY PROOF *",
          " if you want to check them.",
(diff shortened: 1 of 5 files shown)


tirix (approved):
LGTM

PR #142, opened 2023-05-14
Line comments instead of block comments 3

GinoGiotto:
Continuation of https://github.com/metamath/metamath-exe/pull/138.

The same categories of https://github.com/metamath/metamath-exe/pull/139 are applied, although I had to rely on my intuition to make some decisions, so there are a few exceptions to rule 3 and 4 in some circumstances.

PR #144, opened 2023-06-02
Line comments instead of block comments 5

GinoGiotto:
Continuation of https://github.com/metamath/metamath-exe/pull/138.

This PR covers the `mmveri.c` file. Same categories described in https://github.com/metamath/metamath-exe/pull/139 are applied.

I also committed the [third suggestion in the previous PR](https://github.com/metamath/metamath-exe/pull/142#discussion_r1214306901) because Github didn't apply it.

### DIFF
--- a/src/mmveri.c
+++ b/src/mmveri.c
@@ -11,63 +11,62 @@
 #include "mmpars.h"
 #include "mmveri.h"
 
-/* Global structure used for getting information about a step */
-/* If getStep.stepNum is nonzero, we should get info about that step. */
+// Global structure used for getting information about a step.
+// If getStep.stepNum is nonzero, we should get info about that step.
 struct getStep_struct getStep = {0, 0, 0, 0, 0,
     NULL_NMBRSTRING, NULL_NMBRSTRING, NULL_PNTRSTRING, NULL_NMBRSTRING,
     NULL_PNTRSTRING};
 
-/* Verify proof of one statement in source file.  Uses wrkProof structure.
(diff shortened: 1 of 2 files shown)


src/mmveri.c
```diff
@@ -461,7 +454,7 @@ nmbrString *assignVar(nmbrString *bigSubstSchemeAss,
           str(g_Statement[substScheme].numReqHyp),
           " hypotheses.  The ",tmpStr,
           ".  Type \"SHOW PROOF ",g_Statement[statementNum].labelName,
-          "\" to see the proof attempt.",NULL)); */ /* Old version */
+          "\" to see the proof attempt.",NULL)); */ // Old version
```
digama0, line 457: any reason you didn't convert this one?

digama0 (approved):

src/mmveri.c
GinoGiotto: It's a block of commented out code, so it falls in the fourth category described in #139 in case someone would want to use it for experimental use. Do you prefer it to be converted anyway?

digama0: Hm, personally I find it easier to comment out blocks of code using single line comments because there are editor tools to do it, but I'm fine with it either way.

GinoGiotto: Ok, so from now on I won't apply rule 4 of #139 anymore. I'll also make the conversion for the previously edited files, since there were some blocks of code in there as well.

PR #147, opened 2023-06-04
Line comments instead of block comments 8

GinoGiotto:
Continuation of #138 

This PR covers the `mmcmdl.c` file. Only the header and one inline comment are left out.

### DIFF
--- a/src/mmcmdl.c
+++ b/src/mmcmdl.c
@@ -4,7 +4,7 @@
 /*****************************************************************************/
 /*34567890123456 (79-character line to adjust editor window) 2345678901234567*/
 
-/* Command line syntax specification for Metamath */
+// Command line syntax specification for Metamath
 
 #include <string.h>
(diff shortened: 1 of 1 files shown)


src/mmcmdl.c
```diff
@@ -2142,29 +2141,29 @@ long switchPos(const char *swString) {
...
   return 0;
-} /* switchPos */
+} // switchPos
 
 void printCommandError(vstring line1, long arg, vstring errorMsg)
 {
-  /* Warning: errorMsg should not a temporarily allocated string such
-     as the direct output of cat() */
+  // Warning: errorMsg should not a temporarily allocated string such
+  // as the direct output of cat().
```
digama0, line 2149:
```suggestion
  // Warning: errorMsg should not be a temporarily allocated string such
  // as the direct output of cat().
```

PR #148, opened 2023-06-17
Line comments instead of block comments 9

GinoGiotto:
Continuation of https://github.com/metamath/metamath-exe/pull/138

This PR covers the `mmunif.c` file. Only the header, one inline comment and the `/*E*/` comments are left out.

### DIFF
--- a/src/mmunif.c
+++ b/src/mmunif.c
@@ -373,6 +373,7 @@ char unify(
   // Initialize the usage of constants as the first, last,
   // only constant in a $a statement - for rejecting some simple impossible
   // substitutions - Speed-up: this is now done once and never deallocated.
+  //
   // g_firstConst is now cleared in eraseSource.c() (mmcmds.c)
   // to trigger this initialization after "erase".
   if (!nmbrLen(g_firstConst)) {
@@ -514,7 +515,7 @@ char unify(
     }
 
     // Set a flag that the "unknown" variables are not on the stack yet.
-    // (Otherwise this will be the position on the stack).
+    // (Otherwise this will be the position on the stack.)
     for (i = 0; i < unkVarsLen; i++) {
       g_MathToken[unkVars[i]].tmp = -1;
     }
@@ -602,11 +603,11 @@ char unify(
   if (schB[p] > g_mathTokens && schA[p] > g_mathTokens) {
     // Both scheme A and scheme B have variables in the match position.
     // Which one to use?
-    // If neither A nor B is on the stack, use A. Backtrack will put B
-    // on the stack when A's possibilities are exhausted.
-    // If A is on the stack, use A.
-    // If B is on the stack, use B.
-    // If A and B are on the stack, bug.
+    // * If neither A nor B is on the stack, use A. Backtrack will put B
+    //   on the stack when A's possibilities are exhausted.
+    // * If A is on the stack, use A.
+    // * If B is on the stack, use B.
+    // * If A and B are on the stack, bug.
     // In other words:  if B is not on the stack, use A.
     if (g_MathToken[schB[p]].tmp == -1) {
       // B is not on the stack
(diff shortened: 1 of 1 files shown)


src/mmunif.c
```diff
@@ -369,92 +365,92 @@ char unify(
...
 
-  /* Initialize the usage of constants as the first, last,
-     only constant in a $a statement - for rejecting some simple impossible
-     substitutions - Speed-up: this is now done once and never deallocated */
-  /* g_firstConst is now cleared in eraseSource.c() (mmcmds.c)
-     to trigger this initialization after "erase" */
+  // Initialize the usage of constants as the first, last,
+  // only constant in a $a statement - for rejecting some simple impossible
+  // substitutions - Speed-up: this is now done once and never deallocated.
+  // g_firstConst is now cleared in eraseSource.c() (mmcmds.c)
```
digama0, line 376:
```suggestion
  //
  // g_firstConst is now cleared in eraseSource.c() (mmcmds.c)
```
I'm not sure how often to make this suggestion, but there does seem to be some loss of information here where multiple "paragraphs" are running together with the new comment style. Perhaps newlines like this should be used when it would be confusing.

src/mmunif.c
```diff
@@ -504,29 +500,29 @@ char unify(
...
       nmbrLet((nmbrString **)(&stackSaveUnkVarStart[i]),
           stackUnkVar);
       nmbrLet((nmbrString **)(&stackSaveUnkVarLen[i]),
           stackUnkVar);
     }
 
-    /* Set a flag that the "unknown" variables are not on the stack yet */
-    /* (Otherwise this will be the position on the stack) */
+    // Set a flag that the "unknown" variables are not on the stack yet.
+    // (Otherwise this will be the position on the stack).
```
digama0, line 517:
```suggestion
    // (Otherwise this will be the position on the stack.)
```

src/mmunif.c
```diff
@@ -571,73 +566,73 @@ char unify(
...
-    /* If A and B are on the stack, bug. */
-    /* In other words:  if B is not on the stack, use A. */
+    // Both scheme A and scheme B have variables in the match position.
+    // Which one to use?
+    // If neither A nor B is on the stack, use A. Backtrack will put B
+    // on the stack when A's possibilities are exhausted.
+    // If A is on the stack, use A.
+    // If B is on the stack, use B.
+    // If A and B are on the stack, bug.
+    // In other words:  if B is not on the stack, use A.
```
digama0, line 610:
```suggestion
    // * If neither A nor B is on the stack, use A. Backtrack will put B
    //   on the stack when A's possibilities are exhausted.
    // * If A is on the stack, use A.
    // * If B is on the stack, use B.
    // * If A and B are on the stack, bug.
    // In other words:  if B is not on the stack, use A.
```

src/mmunif.c
```diff
@@ -723,125 +718,126 @@ char unify(
...
-    /* We must pop the stack here rather than in backtrack, because we
-       are already one token beyond the end of a scheme, and backtrack
-       would therefore test one token beyond that, missing the fact that
-       the substitution has overflowed beyond the end of a scheme. */
-    /* Set the flag that it's not on the stack and pop stack */
+    // We must pop the stack here rather than in backtrack, because we
+    // are already one token beyond the end of a scheme, and backtrack
+    // would therefore test one token beyond that, missing the fact that
+    // the substitution has overflowed beyond the end of a scheme.
+    // Set the flag that it's not on the stack and pop stack.
```
digama0, line 745:
```suggestion

    // Set the flag that it's not on the stack and pop stack.
```

src/mmunif.c
GinoGiotto: I wondered the same issue while making the changes. My solution was to split comments using periods, but I can see that this might not be enough in some circumstances. I think it's better if I continue with my simple strategy and then you suggest spacings when you think it's appropriate.

PR #151, opened 2023-06-20
Line comments instead of block comments 12

GinoGiotto:
Continuation of https://github.com/metamath/metamath-exe/pull/138

This PR covers the `mmcmds.c` file. Only the header and a few inline comments are left out.

### DIFF
--- a/src/mmcmds.c
+++ b/src/mmcmds.c
@@ -5793,12 +5793,12 @@ void outputMidi(long plen, nmbrString *indentationLevels,
   // read-only input parameters and should not modify the contents of the
   // arrays or strings they point to.
   //
-  // plen = length of proof
-  // indentationLevels[step] = indentation level in "show proof xxx /full"
-  // where step varies from 0 to plen-1.
-  // logicalFlags[step] = 0 for formula-building step, 1 for logical step
-  // g_midiParameter = string passed by user in "midi xxx /parameter <g_midiParameter>"
-  // statementLabel = label of statement whose proof is being scanned.
+  // * plen = length of proof
+  // * indentationLevels[step] = indentation level in "show proof xxx /full"
+  //   where step varies from 0 to plen-1
+  // * logicalFlags[step] = 0 for formula-building step, 1 for logical step
+  // * g_midiParameter = string passed by user in "midi xxx /parameter <g_midiParameter>"
+  // * statementLabel = label of statement whose proof is being scanned
 
   // This function is called when the user types "midi xxx /parameter
   // <g_midiParameter>".  The proof steps of theorem xxx are numbered successively
--- a/src/mmcmds.c
+++ b/src/mmcmds.c
@@ -3697,7 +3697,7 @@ void writeSource(
   flag reformatFlag, // 1 = "/ FORMAT", 2 = "/REWRAP"
   flag splitFlag, // /SPLIT - write out separate $[ $] includes
   flag noVersioningFlag, // /NO_VERSIONING - no ~1 backup
-  // /KEEP_INCLUDES - don't delete included files when /SPIT is not specified
+  // /KEEP_INCLUDES - don't delete included files when /SPLIT is not specified
   flag keepSplitsFlag,
   vstring extractLabelList // "" means /EXTRACT wasn't specified
   )
--- a/src/mmcmds.c
+++ b/src/mmcmds.c
@@ -4277,7 +4277,7 @@ void writeExtractedSource(
           }
         } // if not ${ $}
         if (extractNeeded[stmt + 1] == 'N') {
-// D // printf("added \\n stmt=%ld type=%c,%c\n",stmt+1,g_Statement[stmt].type,g_Statement[stmt+1].type);
+/*D*/ // printf("added \\n stmt=%ld type=%c,%c\n",stmt+1,g_Statement[stmt].type,g_Statement[stmt+1].type);
           // Put a newline following end of statement since the next
           // statement's label section will be suppressed.
           fprintf(fp, "\n");
--- a/src/mmcmds.c
+++ b/src/mmcmds.c
@@ -163,8 +163,7 @@ void typeStatement(long showStmt,
   if (!briefFlag || commentOnlyFlag) {
     free_vstring(str1);
     str1 = getDescription(showStmt);
-    if (!str1[0]) // No comment
-    {
+    if (!str1[0]) { // No comment
       print2("?Warning: Statement \"%s\" has no comment\n",
           g_Statement[showStmt].labelName);
       // We must print a blank comment to have \begin{lemma}
--- a/src/mmcmds.c
+++ b/src/mmcmds.c
@@ -4,63 +4,62 @@
 /*****************************************************************************/
 /*34567890123456 (79-character line to adjust editor window) 2345678901234567*/
 
-/* mmcmds.c - assorted user commands */
+// mmcmds.c - assorted user commands
 
 #include <string.h>
 #include <stdlib.h>
 #include "mmvstr.h"
 #include "mmdata.h"
-#include "mmcmdl.h" /* For g_texFileName */
+#include "mmcmdl.h" // For g_texFileName
 #include "mmcmds.h"
 #include "mminou.h"
 #include "mmpars.h"
 #include "mmveri.h"
-#include "mmwtex.h" /* For g_htmlVarColor,... */
+#include "mmwtex.h" // For g_htmlVarColor,...
 #include "mmpfas.h"
-#include "mmunif.h" /* For g_bracketMatchInit, g_minSubstLen,
-                       ...and g_firstConst */
+#include "mmunif.h" // For g_bracketMatchInit, g_minSubstLen, ...and g_firstConst
 
-/* Local prototypes */
+// Local prototypes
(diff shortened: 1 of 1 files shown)


jkingdon:
At the risk of stating the obvious, `//` comments came in with C99.

Therefore, the #17 task for "Use // comments instead of /* (or doc-comments where appropriate)" is a special case of "No need to worry about versions of the C language older than C99" from https://github.com/metamath/metamath-exe/blob/master/CONTRIBUTING.md so I don't think there is any need to update CONTRIBUTING or the like as part of this set of changes.

digama0:
Well, strictly speaking it's not a special case of the other task which would only support "don't care about `/*` vs `//`" rather than "prefer `//`"

jkingdon:
> Well, strictly speaking it's not a special case of the other task which would only support "don't care about `/*` vs `//`" rather than "prefer `//`" which is what this PR series is implementing.

Strictly speaking, of course that is correct. What the PR series is implementing seems good to me so as far as I can tell we are agreed about what to do.

src/mmcmds.c
```diff
@@ -151,25 +148,26 @@ void typeStatement(long showStmt,
...
-    } /* texFlag */
+    } // texFlag
   }
 
   if (!briefFlag || commentOnlyFlag) {
     free_vstring(str1);
     str1 = getDescription(showStmt);
-    if (!str1[0] /* No comment */) {
+    if (!str1[0]) // No comment
+    {
```
digama0, line 167:
```suggestion
    if (!str1[0]) { // No comment
```

src/mmcmds.c
```diff
@@ -195,48 +193,50 @@ void typeStatement(long showStmt,
...
+    // However, in set.mm the equivalent (and better anyway):
+    //   $d x y ps $.  $d y A $.
+    // produces the same thing when re-merged in SHOW STATEMENT.
     let(&str1, "");
     nmbrTmpPtr1 = g_Statement[showStmt].reqDisjVarsA;
     nmbrTmpPtr2 = g_Statement[showStmt].reqDisjVarsB;
     i = nmbrLen(nmbrTmpPtr1);
-    if (i /* Number of mandatory $d pairs */) {
+    if (i) // Number of mandatory $d pairs
+    {
```
digama0, line 233:
```suggestion
    if (i) { // Number of mandatory $d pairs
```

src/mmcmds.c
```diff
@@ -1259,117 +1254,116 @@ vstring htmlAllowedSubst(long showStmt)
...
-  long splitColumn, /* START_COLUMN */
-  flag skipRepeatedSteps, /* NO_REPEATED_STEPS */
-  flag texFlag,
-  flag htmlFlag
-  /* flag g_midiFlag - global to avoid changing many calls to typeProof() */
+    flag pipFlag, // Means use g_ProofInProgress; statemNum must be proveStatement
+    long startStep, long endStep,
+    long endIndent,
+    // <- also used as definition/axiom flag for HTML syntax breakdown when called
+    // from typeStatement().
```
digama0, line 1276:
```suggestion
    // v- also used as definition/axiom flag for HTML syntax breakdown when called
    // from typeStatement().
```

src/mmcmds.c
```diff
@@ -2729,27 +2726,27 @@ void proofStmtSumm(long statemNum, flag essentialFlag, flag texFlag) {
...
+
+// Traces back the statements used by a proof, recursively.
+// Returns 1 if at least one label is printed (or would be printed in
+// case testOnlyFlag=1); otherwise, returns 0.
+// Returns -1 if the process was aborted.
+// matchList suppresses all output except labels matching matchList.
+// testOnlyFlag prevents any printout; it is used to determine whether
+// there is an unwanted axiom for MINIMIZE_WITH /FORBID.
+// abortOnQuit is set to 1 if we want to abort early if the user selects
+// 'Q to quit' at the prompt, and if so we return -1.
```
digama0, line 2749:
```suggestion
// * matchList suppresses all output except labels matching matchList.
// * testOnlyFlag prevents any printout; it is used to determine whether
//   there is an unwanted axiom for MINIMIZE_WITH /FORBID.
// * abortOnQuit is set to 1 if we want to abort early if the user selects
//   'Q to quit' at the prompt, and if so we return -1.
```

src/mmcmds.c
```diff
@@ -3437,145 +3434,145 @@ vstring bigAdd(vstring bignum1, vstring bignum2) {
...
+// depending on whether g_Statement[i] uses statemNum.  All i will be populated
+// with 'Y'/'N' even if not $a or $p (always 'N' for non-$a,$p).
+// Optional 'cutoffStmt' parameter:  if nonzero, then
+// statements above cutoffStmt will not be scanned (for speedup).
 vstring traceUsage(long statemNum,
   flag recursiveFlag,
-  long cutoffStmt /* for speedup */) {
-
+  long cutoffStmt) // for speedup
+  {
```
digama0, line 3564:
```suggestion
  long cutoffStmt) // for speedup
{
```

src/mmcmds.c
```diff
@@ -3693,67 +3689,67 @@ void readInput(void)
...
-  flag reformatFlag /* 1 = "/ FORMAT", 2 = "/REWRAP" */,
-  flag splitFlag,  /* /SPLIT - write out separate $[ $] includes */
-  flag noVersioningFlag, /* /NO_VERSIONING - no ~1 backup */
-  flag keepSplitsFlag, /* /KEEP_INCLUDES - don't delete included
-                        files when /SPIT is not specified */
-  vstring extractLabelList /* "" means /EXTRACT wasn't specified */
+  flag reformatFlag, // 1 = "/ FORMAT", 2 = "/REWRAP"
+  flag splitFlag, // /SPLIT - write out separate $[ $] includes
+  flag noVersioningFlag, // /NO_VERSIONING - no ~1 backup
+  // /KEEP_INCLUDES - don't delete included files when /SPIT is not specified
```
digama0, line 3702:
```suggestion
  // /KEEP_INCLUDES - don't delete included files when /SPLIT is not specified
```

src/mmcmds.c
```diff
@@ -4244,56 +4240,56 @@ void writeExtractedSource(
...
-          /* $v $c $d $e $f $a $p */
+          // $v $c $d $e $f $a $p
           let(&buf, space(g_Statement[stmt].mathSectionLen));
           memcpy(buf, g_Statement[stmt].mathSectionPtr,
               (size_t)(g_Statement[stmt].mathSectionLen));
           fprintf(fp, "%s", buf);
           if (g_Statement[stmt].type != p_) {
             fprintf(fp, "$.");
-/*D*//*fprintf(fp, "#%ld#",stmt);*/
+// D // fprintf(fp, "#%ld#",stmt);
```
digama0, line 4271:
```suggestion
/*D*/ // fprintf(fp, "#%ld#",stmt);
```
I forget what style exactly you were using for these but you should keep it consistent

src/mmcmds.c
```diff
@@ -4244,56 +4240,56 @@ void writeExtractedSource(
...
                 (size_t)(g_Statement[stmt].proofSectionLen));
             fprintf(fp, "%s$.", buf);
           }
-        } /* if not ${ $} */
+        } // if not ${ $}
         if (extractNeeded[stmt + 1] == 'N') {
-/*D*//*printf("added \\n stmt=%ld type=%c,%c\n",stmt+1,g_Statement[stmt].type,g_Statement[stmt+1].type);*/
-          /* Put a newline following end of statement since the next
-             statement's label section will be suppressed */
+// D // printf("added \\n stmt=%ld type=%c,%c\n",stmt+1,g_Statement[stmt].type,g_Statement[stmt+1].type);
```
digama0, line 4282:
```suggestion
/*D*/ // printf("added \\n stmt=%ld type=%c,%c\n",stmt+1,g_Statement[stmt].type,g_Statement[stmt+1].type);
```

src/mmcmds.c
```diff
@@ -5772,98 +5773,98 @@ long getStatementNum(vstring stmtName, /* Possibly with wildcards */
...
+  // The parameters have the following meanings.  You should treat them as
+  // read-only input parameters and should not modify the contents of the
+  // arrays or strings they point to.
+  //
+  // plen = length of proof
+  // indentationLevels[step] = indentation level in "show proof xxx /full"
+  // where step varies from 0 to plen-1.
+  // logicalFlags[step] = 0 for formula-building step, 1 for logical step
+  // g_midiParameter = string passed by user in "midi xxx /parameter <g_midiParameter>"
+  // statementLabel = label of statement whose proof is being scanned.
```
digama0, line 5803:
```suggestion
  // * plen = length of proof
  // * indentationLevels[step] = indentation level in "show proof xxx /full"
  //   where step varies from 0 to plen-1
  // * logicalFlags[step] = 0 for formula-building step, 1 for logical step
  // * g_midiParameter = string passed by user in "midi xxx /parameter <g_midiParameter>"
  // * statementLabel = label of statement whose proof is being scanned
```

digama0 (approved):

src/mmcmds.c
GinoGiotto: Ah yes, I left all `/*E*/` comments in the original form, I must have missed that this `/*D*/` comment was of the same kind.

Issue #129, opened 2023-01-25
Latex curly brackets bug


GinoGiotto:
Hello, so I wanted to show a proof in a .tex file, to do this I executed the following commands in Metamath:

<pre>Metamath - Version 0.198 7-Aug-2021           Type HELP for help, EXIT to exit.
MM> read set.mm
Reading source file "set.mm"... 42839696 bytes
42839696 bytes were read into the source buffer.
The source has 203558 statements; 2716 are $a and 40625 are $p.
No errors were found.  However, proofs were not checked.  Type VERIFY PROOF *
if you want to check them.
MM> VERIFY PROOF *
0 10%  20%  30%  40%  50%  60%  70%  80%  90% 100%
..................................................
All proofs in the database were verified in 27.72 s.
MM> open tex proof-example.tex
Created LaTeX output file "proof-example.tex".
Reading definitions from $t statement of set.mm...
1643 typesetting statements were read from "set.mm".
MM> show proof rankidn /tex
Outputting proof of "rankidn"...
The LaTeX source was written to "proof-example.tex".
MM> close tex
The LaTeX output file "proof-example.tex" has been closed.
MM>
</pre>

These commands create [THIS](https://github.com/metamath/metamath-exe/files/10494988/proof-example.txt) file called "proof-example.tex" (I changed the format to .txt) containing the proof of the "rankidn" theorem.

Now, when I compile the .tex file with a latex editor (I tested it with TeXstudio, TeXworks and Texmaker) it shows some errors such as:
<pre>
Missing } inserted. \end{align}
Missing { inserted. \end{align}
Missing \endgroup inserted. \end{align}
Misplaced \omit. \end{align}
</pre>

It also shows some warnings regarding missing references (latex commands such as <code>\ref{eq:rankr1c}</code> mention theorems and axioms that are not present in the .tex file so it shows <code>(??)</code> in the generated pdf). This is more of an aesthetic issue and it's not the main theme of this bug report.

Now regarding the errors here what I think is happening:

I think they all have one single cause, and the cause is that curly brackets cannot be opened and closed in separate lines. More specifically none of these symbols: <code> \\\ </code> <code> && </code> <code> & </code> should appear in some text enclosed in curly brackets.

It's possible to test my hypothesis with the example I provided: If you edit the second step of "proof-example.tex" from:
<pre>
\\ 2 &&  & \vdash ( A \in \bigcup ( R_1 `` {\rm On} ) \rightarrow ( ( {\rm
[...]

PR #166, opened 2023-10-22
Attempting to fix the curly brackets bug

GinoGiotto:
This is my attempt to fix issue #129. The solution I proposed in https://github.com/metamath/metamath-exe/issues/129#issuecomment-1404255591 is actually not very practical because latex translations get merged togheter pretty early and this makes it difficult to take advantage of that information. So I changed my mind and implemented @david-a-wheeler proposal instead https://github.com/metamath/metamath-exe/issues/129#issuecomment-1435743413. I took inspiration from an analogous functionality already present for HTML: https://github.com/metamath/metamath-exe/blob/ae35c1b2931be8f9683f8bf66d1d10c9b56db706/src/mminou.c#L549-L588

With this change the generated TeX of my [minimal example](https://github.com/metamath/metamath-exe/issues/129#issuecomment-1765135611) seems to work as intended:
<pre>
% This LaTeX file was created by Metamath on 22-Oct-2023 9:40 PM.
\documentclass{article}
\usepackage{amssymb} % amssymb must be loaded before phonetic
\usepackage{phonetic} % for \riota
\usepackage{mathrsfs} % for \mathscr
\usepackage{mathtools} % loads package amsmath
\usepackage{amsthm} % amsthm must be loaded after amsmath
\usepackage{accents} % accents should be loaded after mathtools
\theoremstyle{plain}
\newtheorem{theorem}{Theorem}[section]
\newtheorem{definition}[theorem]{Definition}
\newtheorem{lemma}[theorem]{Lemma}
\newtheorem{axiom}{Axiom}
\allowdisplaybreaks[1] % Allow page breaks in {align}
\usepackage[plainpages=false,pdfpagelabels]{hyperref}
\hypersetup{colorlinks} % Get rid of boxes around links
\begin{document}

\begin{proof}
\begin{align}
  1 &&  & \text{ A A A A A A A A A A } \text{ A A A A A A A A A A }
    \notag \\ && & \qquad \text{ A A A A A A A A A A }
    \notag \\ && & \qquad \text{ A A A A A A A A A A }
\text{ A A A A A A A A A A }&(\mbox{\ref{eq:ax-1}})\notag
\end{align}
\end{proof}

\end{document}
</pre>

I'm somewhat surprised by how simple this solution looks, and this makes me wonder whether it might be a false friend due to my own ignorance. So it most likely needs to be overhauled by an expert eye.

### DIFF
--- a/src/mminou.c
+++ b/src/mminou.c
@@ -501,7 +501,7 @@ void printLongLine(const char *line, const char *startNextLine, const char *brea
   vstring_def(prefix);
   vstring_def(startNextLine1);
   vstring_def(breakMatch1);
-  long i, j, p, k;
+  long i, p;
   long startNextLineLen;
   flag firstLine;
   flag tildeFlag = 0;
@@ -546,12 +546,6 @@ void printLongLine(const char *line, const char *startNextLine, const char *brea
     breakMatch1[0] = ' '; // Change to a space (the real break character)
   }
 
-  // Do a bug check to make sure no real ASCII 3's are ever printed
-  j = (long)strlen(multiLine);
-  for (i = 0; i < j; i++) {
-    if (multiLine[i] == QUOTED_SPACE) bug(1514); // Should never be the case
-  }
-
   // HTML mode
   // The HTML mode is intended not to break inside quoted HTML tag
   // strings.  All HTML output should be called with this mode.
@@ -587,26 +581,6 @@ void printLongLine(const char *line, const char *startNextLine, const char *brea
       i++;
     }
   } // if (breakMatch1[0] == '\"')
-
-  // TeX mode
-  // The TeX mode is intended not to break inside curly brace scopes.
-  // Whenever we are inside a scope of curly braces, we change a space to
-  // ASCII 3 to prevent matching it.  The reverse is done in the print2()
-  // function, where all ASCII 3's are converted back to space.
-  if (breakMatch1[0] == ' ') {
-    i = 0;
-    // k counts the scope level we are in.
-    k = 0;
-    while (multiLine[i]) {
-      // We enter a non "\{" scope.
-      if (multiLine[i] == '{' && multiLine[i - 1] != '\\') k++;
-      // We escape a non "\}" scope.
-      if (multiLine[i] == '}' && multiLine[i - 1] != '\\') k--;
-      // If k > 0 then we are inside a scope.
-      if (multiLine[i] == ' ' && k > 0) multiLine[i] = QUOTED_SPACE;
-      i++;
-    }
-  }
 
   // The tilde is a special flag for printLongLine to print a
   // tilde before the carriage return in a split line, not after.
--- a/src/mminou.c
+++ b/src/mminou.c
@@ -501,7 +501,7 @@ void printLongLine(const char *line, const char *startNextLine, const char *brea
   vstring_def(prefix);
   vstring_def(startNextLine1);
   vstring_def(breakMatch1);
-  long i, j, p;
+  long i, j, p, k;
   long startNextLineLen;
   flag firstLine;
   flag tildeFlag = 0;
@@ -546,6 +546,12 @@ void printLongLine(const char *line, const char *startNextLine, const char *brea
     breakMatch1[0] = ' '; // Change to a space (the real break character)
   }
 
+  // Do a bug check to make sure no real ASCII 3's are ever printed
+  j = (long)strlen(multiLine);
+  for (i = 0; i < j; i++) {
+    if (multiLine[i] == QUOTED_SPACE) bug(1514); // Should never be the case
+  }
+
   // HTML mode
   // The HTML mode is intended not to break inside quoted HTML tag
   // strings.  All HTML output should be called with this mode.
@@ -559,11 +565,6 @@ void printLongLine(const char *line, const char *startNextLine, const char *brea
   // where all ASCII 3's are converted back to space.
   // Note added 20-Oct-02: tidy.exe breaks HREF quotes with new line.
   // Check HTML spec - do we really need this code?
-  j = (long)strlen(multiLine);
-  // Do a bug check to make sure no real ASCII 3's are ever printed
-  for (i = 0; i < j; i++) {
-    if (multiLine[i] == QUOTED_SPACE) bug(1514); // Should never be the case
-  }
   if (breakMatch1[0] == '\"') {
     breakMatch1[0] = ' '; // Change to a space (the real break character)
     // Scan string for quoted strings
@@ -586,6 +587,26 @@ void printLongLine(const char *line, const char *startNextLine, const char *brea
       i++;
     }
   } // if (breakMatch1[0] == '\"')
+
+  // TeX mode
+  // The TeX mode is intended not to break inside curly brace scopes.
+  // Whenever we are inside a scope of curly braces, we change a space to
+  // ASCII 3 to prevent matching it.  The reverse is done in the print2()
+  // function, where all ASCII 3's are converted back to space.
+  if (breakMatch1[0] == ' ') {
+    i = 0;
+    // k counts the scope level we are in.
+    k = 0;
+    while (multiLine[i]) {
+      // We enter a non "\{" scope.
+      if (multiLine[i] == '{' && multiLine[i - 1] != '\\') k++;
+      // We escape a non "\}" scope.
+      if (multiLine[i] == '}' && multiLine[i - 1] != '\\') k--;
+      // If k > 0 then we are inside a scope.
+      if (multiLine[i] == ' ' && k > 0) multiLine[i] = QUOTED_SPACE;
+      i++;
+    }
+  }
 
   // The tilde is a special flag for printLongLine to print a
   // tilde before the carriage return in a split line, not after.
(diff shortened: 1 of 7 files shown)


src/mminou.c
```diff
@@ -587,6 +588,26 @@ void printLongLine(const char *line, const char *startNextLine, const char *brea
...
+      // We enter a non "\{" scope.
+      if (multiLine[i] == '{' && multiLine[i - 1] != '\\') k++;
+      // We escape a non "\}" scope.
+      if (multiLine[i] == '}' && multiLine[i - 1] != '\\') k--;
+      // If k > 0 then we are inside a scope.
+      if (multiLine[i] == ' ' && k > 0) multiLine[i] = QUOTED_SPACE;
+      i++;
+    }
+  }
+
```
tirix, line 610: The function is `printLongLine`, I think it's used to print out pretty much everything.
I don't see any check that we actually are in TeX mode. Is this code inoffensive in e.g. ASCII mode?

GinoGiotto: You're right. 

So, to address this I'm thinking about imitating the HTML again. The curly braces issue affects only these two printed lines:

https://github.com/metamath/metamath-exe/blob/ae35c1b2931be8f9683f8bf66d1d10c9b56db706/src/mmwtex.c#L3046-L3099

So the idea would be to change `breakMatch` (the third argument of `printLongLine`) from a space to a special character, to be sure that we are in the above cases only. Then we change the special charater to a space again so that the following rewrapping algorithm works as expected:
https://github.com/metamath/metamath-exe/blob/ae35c1b2931be8f9683f8bf66d1d10c9b56db706/src/mminou.c#L567-L568

What do you think?

tirix: What about only performing that treatment within this condition?
```
if (!g_htmlFlag) { // LaTeX 
   ...
}
```
it's supposed to be a global flag switching the output mode between HTML and LaTeX.
What I'm afraid of, is that it might still be set when we print to the standard output.

GinoGiotto: You mean converting all spaces into `QUOTED_SPACE` inside the `if (!g_htmlFlag) { // LaTeX` condition instead of the mminou.c file? That might be not a bad idea.

tirix: To be honest I'm not quite sure, I certainly did not take time to fully analyze the code...

GinoGiotto: I attempted to put the brace count inside the `if (!g_htmlFlag) { // LaTeX` condition and it seems to work. But before eventually committing it I have a question:

With this new apporach the `line` argument of `printLongLine` contains some `QUOTED_SPACE` characters and that triggers bug check 1514. The reason why it gets triggered it pretty straightforward since `bug(1514)` checks exacly that:

https://github.com/metamath/metamath-exe/blob/ae35c1b2931be8f9683f8bf66d1d10c9b56db706/src/mminou.c#L562-L566

I don't understand the point of this bug check, it seems that its purpose is to make sure that no `QUOTED_SPACE` is printed, but, in my eyes it fails at checking what it claims to check. According to the `// HTML mode` comment, all occurrences of  `QUOTED_SPACE` get converted back into spaces inside the `print2()` function, so wouldn't be more useful to put this check in `print2()` instead? Maybe at the end of all operations, right before printing?

So to make it work I had to remove the above lines and the resulting TeX looks fine.

tirix: What I understand is the following:
The algorithm wants to prevent line breaking at certain places (spaces within double-quote `"` quotation). To do so, it first replaces all such spaces by `QUOTED_SPACE`, then performs the line breaking, which acts on the remaining spaces, and finally switches back the `QUOTED_SPACE ` to normal spaces. 

This is supposed to be transparent, in the sense that space in the input results in spaces or line breaks in the output, and no spaces are added.

If there are `QUOTED_SPACE` in the original string, even before the whole process, those would be turned into normal spaces at the end of the processing, so spaces are "added" and the above contract is not respected. This makes `QUOTED_SPACE` a special character which should not be present in the input (and therefore the bug 1514).

Here, your (our) changes would expose some inner working of `printLongLine` to the caller of the function, and its contract becomes:
- the function breaks long lines into suitable mutli-lines according to the rules, etc., as before,
- any `QUOTED_SPACE` character is replaced by a space in the output, and cannot be a line break position.

We're using this "new" functionality in the caller by putting `QUOTED_SPACE` where LaTeX does not want line breaks to appear.
And because the `QUOTED_SPACE` thing is now part of the function's pre/post conditions, it is not a reserved character anymore and we can remove the bug check safely.

If this is agreed and we move on with the change, it would be nice to put some of these thoughts in or around the function's comment!

GinoGiotto: Yep, everything you said aligns with my thoughts, so it seems we are on the same page. Unless somebody else have objections/other proposals, I'm going to commit the new proposed solution soon (it basically subsumes the old one).

GinoGiotto: I committed the new proposal. To keep topics in order, I mark this thread as resolved since we reached an agreement. Open a new one for further discussion.

src/mmwtex.c
```diff
@@ -3047,55 +3057,94 @@ void printTexLongMath(nmbrString *mathString,
...
+        // The spaces that should not be matched with 'breakMatch' are
+        // temporarily changed to ASCII 3 before the 'printLongLine' procedure
+        // is called.  The procedure rewraps the 'line' argument matching the
+        // unchanged spaces only, thus ensuring bad breaks will be avoided.
+        // The reverse is done in the 'print2()' function, where all ASCII 3's
+        // are converted back to spaces.
+        // k counts the scope level we are in.
+        k = 0;
+        n = (long)strlen(texFull);
+        for (j = 1; j < n; j++) { // We don't need to check texFull[0].
```
GinoGiotto, line 3089: Here I wasn't sure what apporach would be best. I defined the `texFull` string as the line that willl be feeded to the `printLongLine` function. The checking cicle starts at the second character of `texFull` to avoid checking the unorthodox `texFull[-1] != '\\'`. The first character is `texFull[0] = " "` so we don't need to check it since it's fixed.

src/mmwtex.c
```diff
@@ -3047,55 +3057,94 @@ void printTexLongMath(nmbrString *mathString,
...
-            // The following puts a hypothesis number such as "2" if
-            // $e label is "abc.2"; if no ".", will be whole label.
-            right(htmRef, instr(1, htmRef, ".") + 1),
-            "}\\notag%",
-            // Add full label as LaTeX comment - note lack of space after
-            // "%" above to prevent bad line break.
-            htmRef, NULL),
-            "    \\notag \\\\ && & \\qquad ", // Continuation line prefix
-            " ");
+        texFull = cat("  ", // texFull[0] should not be a "{" character.
```
GinoGiotto, line 3060: Here I added a comment to point out that `texFull[0] = " "` should not be a "{" character, since it wouldn't be checked by the subsequent counting algorithm.

tirix: Ok, and here it obviously will never be "`{`", since it is a blank space "` `".

src/mmwtex.c
tirix: Looks good.

src/mmwtex.c
```diff
@@ -3047,55 +3057,94 @@ void printTexLongMath(nmbrString *mathString,
...
+        n = (long)strlen(texFull);
+        for (j = 1; j < n; j++) { // We don't need to check texFull[0].
+          // We enter a non "\{" scope.
+          if (texFull[j] == '{' && texFull[j - 1] != '\\') k++;
+          // We escape a non "\}" scope.
+          if (texFull[j] == '}' && texFull[j - 1] != '\\') k--;
+          // If k > 0 then we are inside a scope.
+          if (texFull[j] == ' ' && k > 0) texFull[j] = QUOTED_SPACE;
+        }
+        printLongLine(texFull, "    \\notag \\\\ && & \\qquad ", /* Continuation line prefix */ " ");
```
tirix, line 3147: Looks like a lot of duplication with the previous part.
Then again the duplicated code was already there before.

tirix (approved):

src/mmwtex.c
GinoGiotto: I unduplicated the brace count and the `printLongLine` call in a new commit.

tirix:
Maybe one (last?) remark would be to integrate your minimal example into an additional test in the `tests` directory, with an additional `issue129.in`, `issue129.mm` and `issue129.expected`.

Unfortunately I don't have the rights to merge, I guess @digama0 does.

src/mmwtex.c
tirix: Looks good!

GinoGiotto:
My only question is whether `issue129.tex` should be included as well, since in this case that's the file we are interested verifying. But `run_test.sh` doesn't seem to be designed to check the generated tex tho.

tirix:
Oh right, of course I have not thought so far. 
That would have been a nice to have, I'm sorry I feel I have made you add a quite useless test.

digama0:
There is another test which has to check the result of an auxiliary file, and uses a trick calling `cat` inside metamath to get it to be validated. See the `underscores` test.

GinoGiotto:
I don't understand what's wrong. On my computer both attempts pass. Maybe it doesn't like backslashes?

tirix:
Based on [this SO answer](https://stackoverflow.com/questions/43528202/prevent-echo-from-interpreting-backslash-escapes), line 113 of `run_tests.sh`, I would try replacing the `cat "$result"` with `printf '%s' "$result"` so that the backslashes are not interpreted.

GinoGiotto:
Thanks for making it work, should be good to merge now.

I wonder if there is any diff

GinoGiotto:
Btw sorry @tirix I superficially read your suggestion and made a mess.

tirix:
The final one looks good!
I already approved, I think only @digama0 can merge now.

digama0 (approved):

PR #167, opened 2023-10-29
mmwtex.c: boolean-style evaluation in conditionals (+ linting)

benjub:
Found some inconsistencies while on #160, which I signal as TODOs in this PR. I also made some simplifications (mainly, evaluate boolean-like variables as bools) and linting, all in `mmwtex.c`.

### DIFF
--- a/src/mmwtex.c
+++ b/src/mmwtex.c
@@ -25,16 +25,15 @@
 flag g_oldTexFlag = 0; // Use TeX macros in output (obsolete)
 
 flag g_htmlFlag = 0; // HTML flag: 0 = TeX, 1 = HTML
+
 // Use "althtmldef" instead of "htmldef".  This is intended to allow the
-// generation of pages with the Unicode font instead of the individual GIF files.
+// generation of pages with the Unicode font instead of the individual GIF
+// files.
 flag g_altHtmlFlag = 0;
-// Output statement lists only, for statement display in other HTML pages,
-// such as the Proof Explorer home page.
-flag g_briefHtmlFlag = 0; // TODO: unused
-// At this statement and above, use the exthtmlxxx
-// variables for title, links, etc.  This was put in to allow proper
-// generation of the Hilbert Space Explorer extension to the set.mm
-// database.
+
+// At this statement and above, use the exthtmlxxx variables for title, links,
+// etc.  This was put in to allow proper generation of the Hilbert Space
+// Explorer extension to the set.mm
 long g_extHtmlStmt = 0;
 
 // Globals to hold mathbox information.  They should be re-initialized
--- a/src/mmwtex.c
+++ b/src/mmwtex.c
@@ -67,7 +67,7 @@ pntrString_def(g_mathboxUser); // User name vs. mathbox number
  */
 #define QUOTED_SPACE 3 // ASCII 3 that temporarily zaps a space
 
-// LaTex output file
+// LaTeX output file
 FILE *g_texFilePtr = NULL;
 flag g_texFileOpenFlag = 0;
 
--- a/src/mmwtex.c
+++ b/src/mmwtex.c
@@ -4834,7 +4834,7 @@ vstring spectrumToRGB(long color, long maxColor) {
   return str1;
 } // spectrumToRGB
 
-// Return the HTML code for GIFs (!g_altHtmlFlag) or Unicode (g_altHtmlFlag),
+// Returns the HTML code for GIFs (!g_altHtmlFlag) or Unicode (g_altHtmlFlag),
 // or LaTeX when !g_htmlFlag, for the math string (hypothesis or conclusion) that
 // is passed in.
 // Warning: The caller must deallocate the returned vstring.
--- a/src/mmwtex.c
+++ b/src/mmwtex.c
@@ -1819,7 +1819,6 @@ flag printTexComment(vstring commentPtr, flag htmlCenterFlag,
     let(&cmtMasked, tmpMasked);
     free_vstring(tmpStr); // Deallocate
     free_vstring(tmpStrMasked);
-  }
 
     // Add leading and trailing HTML markup to comment here
     // (instead of in caller).  Also convert special characters.
--- a/src/mmwtex.c
+++ b/src/mmwtex.c
@@ -852,7 +852,7 @@ long texDefWhiteSpaceLen(char *ptr)
     i++;
   }
   bug(2307);
-  assert(0);
+  return 0;
 } // texDefWhiteSpaceLen
 
 // This function returns the length of the token (non-whitespace) starting at
@@ -894,7 +894,7 @@ long texDefTokenLen(char *ptr)
     i++;
   }
   bug(2308);
-  assert(0);
+  return 0;
 } // texDefTokenLen
 
 // Token comparison for qsort
@@ -1222,7 +1222,6 @@ vstring getCommentModeSection(vstring *srcptr, char *mode)
     }
     ptr++;
   } // while (1)
-  assert(0);
 } // getCommentModeSection
 
 // The texHeaderFlag means this:
--- a/src/mmwtex.c
+++ b/src/mmwtex.c
@@ -1649,7 +1649,7 @@ void printTexHeader(flag texHeaderFlag)
   free_vstring(tmpStr);
 } // printTexHeader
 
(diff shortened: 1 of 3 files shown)


src/mmwtex.c
```diff
@@ -30,7 +30,7 @@ flag g_htmlFlag = 0; // HTML flag: 0 = TeX, 1 = HTML
 flag g_altHtmlFlag = 0;
 // Output statement lists only, for statement display in other HTML pages,
 // such as the Proof Explorer home page.
-flag g_briefHtmlFlag = 0;
+flag g_briefHtmlFlag = 0; // TODO: unused
```
tirix, line 33: Unused? What about https://github.com/metamath/metamath-exe/blob/ae35c1b2931be8f9683f8bf66d1d10c9b56db706/src/metamath.c#L2854
And https://github.com/metamath/metamath-exe/blob/ae35c1b2931be8f9683f8bf66d1d10c9b56db706/src/metamath.c#L2616

digama0: And
https://github.com/metamath/metamath-exe/blob/ae35c1b2931be8f9683f8bf66d1d10c9b56db706/src/metamath.c#L2578-L2587
which shows that it is not a constant either.

src/mmwtex.c
```diff
@@ -59,7 +59,7 @@ pntrString_def(g_mathboxUser); // User name vs. mathbox #
 // in a math string in a comment to be removed for HTML output.
 #define CLOSING_PUNCTUATION ".,;)?!:]'\"_-"
 
-// Tex output file
+// LaTeX output file
```
digama0, line 62: nitpick: Does metamath-exe actually produce LaTeX (as compared to TeX) output? I would not be surprised if it is plain TeX.

src/mmwtex.c
```diff
@@ -853,10 +853,10 @@ long texDefWhiteSpaceLen(char *ptr)
     i++;
   }
   bug(2307);
-  return 0; // Dummy return - never executed
+  assert(0);
```
digama0, line 856: This is incorrect, `bug` can return! The user is provided a prompt and one of the options is to continue.

src/mmwtex.c
benjub: Should I move the declaration/initialization `flag g_briefHtmlFlag = 0;` to `metamath.c` ?

src/mmwtex.c
benjub: So, I restore `return 0;` but remove its comment ?

src/mmwtex.c
benjub: `metamath-exe` produces LaTeX code that is not TeX-compatible, since it outputs "\documentclass" for instance, and uses several packages.

src/mmwtex.c
digama0: I think it could even be moved to a local variable in `command()`, one less global variable to worry about.

benjub: done in 6bc25466c16958979ad8ab106381c3049d086606

src/mmwtex.c
benjub: done in 811908c3ef9857018473e9d66022cda45bc0e6dd

src/mmwtex.c
```diff
@@ -1222,7 +1222,6 @@ vstring getCommentModeSection(vstring *srcptr, char *mode)
     }
     ptr++;
   } // while (1)
-  assert(0);
```
benjub: Am I correct ? This follows a `while (1)`  loop with no `break;` statement in the outermost loop (the `break;` is in a `switch` statement within the `while(1)` loop). Therefore, this is unreachable code.

digama0: The `break;` after `return modeSection;` looks very strange in that code, that is also clearly unreachable code.

benjub: Done in 6523fc7c1a292fdbba145b11d65a5d2f876e518f
I didn't go further and didn't change the `switch` statement since I don't know the code enough (`mode` seems to take "magic" values, for instance).

tirix (approved):

benjub:
@digama0 : following approval by Thierry, can you merge this one ?

src/mmwtex.c
```diff
@@ -16,25 +16,24 @@
...
-// such as the Proof Explorer home page.
-flag g_briefHtmlFlag = 0;
-// At this statement and above, use the exthtmlxxx
-// variables for title, links, etc.  This was put in to allow proper
-// generation of the Hilbert Space Explorer extension to the set.mm
-// database.
+
+// At this statement and above, use the exthtmlxxx variables for title, links,
+// etc.  This was put in to allow proper generation of the Hilbert Space
+// Explorer extension to the set.mm
```
digama0, line 36:
```suggestion
// Explorer extension to the set.mm database.
```

src/mmwtex.c
```diff
@@ -68,7 +67,7 @@ pntrString_def(g_mathboxUser); // User name vs. mathbox #
  */
 #define QUOTED_SPACE 3 // ASCII 3 that temporarily zaps a space
 
-// Tex output file
+// LaTex output file
```
digama0, line 70:
```suggestion
// LaTeX output file
```

src/mmwtex.c
```diff
@@ -197,7 +196,7 @@ flag readTexDefs(
     eraseTexDefs();
     saveHtmlFlag = g_htmlFlag; // Save for next call to readTexDefs()
     saveAltHtmlFlag = g_altHtmlFlag; // Save for next call to readTexDefs()
-    if (g_htmlFlag == 0 /* Tex */ && g_altHtmlFlag == 1) {
+    if (!g_htmlFlag /* LaTex */ && g_altHtmlFlag) {
```
digama0, line 199:
```suggestion
    if (!g_htmlFlag /* LaTeX */ && g_altHtmlFlag) {
```

src/mmwtex.c
```diff
@@ -1086,15 +1086,16 @@ vstring tokenToTex(vstring mtoken, long statemNum /* for error msgs */)
...
-  } // End if
+    }
+  } // if (texDefsPtr)
 
   return tex;
 } // tokenToTex
 
-// Converts a comment section in math mode to TeX.  Each math token
-// MUST be separated by white space.   TeX "$" does not surround the output.
+// Convert a comment section in math mode to LaTeX.  Each math token
```
digama0, line 1097:
```suggestion
// Converts a comment section in math mode to LaTeX.  Each math token
```
I think indicative mood is more conventional for documentation and is already used elsewhere to my knowledge

src/mmwtex.c
```diff
@@ -1651,10 +1649,10 @@ void printTexHeader(flag texHeaderFlag)
   free_vstring(tmpStr);
 } // printTexHeader
 
-// Prints an embedded comment in TeX or HTML.  The commentPtr must point to the first
+// Print an embedded comment in LaTeX or HTML.  The commentPtr must point to the first
```
digama0, line 1652:
```suggestion
// Prints an embedded comment in LaTeX or HTML.  The commentPtr must point to the first
```

src/mmwtex.c
```diff
@@ -4658,7 +4653,7 @@ flag getSectionHeadings(long stmt,
   return errorFound;
 } // getSectionHeadings
 
-// Returns HTML for the pink number to print after the statement labels
+// Return HTML for the pink number to print after the statement labels
```
digama0, line 4656:
```suggestion
// Returns HTML for the pink number to print after the statement labels
```

digama0 (approved):

benjub:
I didn't apply the four suggestions changing imperative to indicative since I thought it was the opposite convention, but I'm unsure. I'll let you decide to apply or not before merging.

digama0:
I took a census of how this is used in the .h file comments on all functions. As expected, there has not historically been any consistency about this, but the indicatives seem to be winning:

* Imperative mood: 73 functions
* Indicative mood: 94 functions
* Noun phrase: 31 functions

Doxygen seems to have thrown its opinion into the ring slightly with the `\returns` keyword. My preference is to standardize on indicative mood, perhaps it should be added as an additional element to the "code cleanup" issue?

benjub:
> perhaps it should be added as an additional element to the "code cleanup" issue?

Done. (Add suggestions applied.)

PR #177, opened 2024-09-18
fix README.TXT

wlammen:
After recovering from my accident, and buying a replacement for my broken notebook, I started installing Metamath on it.  One need is to have a metamath exe available, and I compiled it from scratch.

I use an Linux/Ubuntu OS, and not everything went according to plan.  Besides the corrections in this PR, I stumbled over following issues:
1. Current Ubuntu implementations provide a gcc C compiler for 64 bit code only.  The autoconf procedure somehow requires an additional ia32 implementation, which you get by installing gcc-multilib
sudo apt-get install gcc-multilib
2. The rlwrap home page moved.  To compile it you need to have readline and curses
sudo apt-get install libreadline-dev
sudo apt-get install libncurses5-dev libncursesw5-dev

### DIFF
--- a/README.TXT
+++ b/README.TXT
@@ -37,6 +37,12 @@ set.mm locally (cp /usr/share/metamath/set.mm . ; metamath set.mm), or run
 metamath and type:  read "/usr/share/metamath/set.mm" (note that inside
 metamath, filenames containing "/" must be quoted).
 
+Known issues with the autoconf approach:  On Intel type processors x86 the
+configure script might want you to support 32-bit code, even if your system
+is natively 64-bit.  This is known as cross-compiling and on Debian you need
+the package gcc-multilib installed.  For other Linux OS a similar extension
+might be in order.
+
 
 Optional enhancements
 ---------------------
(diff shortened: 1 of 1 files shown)


digama0:
Did you want to add (1) and (2) to the readme as well?

wlammen:
(1) Adding a known issues / best practices paragraph makes sense if it is not just me who spent extra time to find the final solution-
(2) I did not expect rlwrap be available as a debian package, but it is!  So of course `sudo apt-get install rlwrap` is sufficient.

digama0:
The readme is for people trying to set it up, so it makes sense that if you had trouble following the directions then you should add enough information so that the next person won't.

wlammen:
This is a can of worms.  Debian packages are only used on Debian installations.  The autoconf process addresses a wider range of OS including main frames and exotic Linux variants, and of all machines only Intel processors seem to need the gcc extension.  This is only the beginning.
I don't have more than anecdotal knowledge here.  It is not my intention to study the gory details for a profound explanation.  What I do here is correcting obvious errors, and as little as this seems, it is an improvement.
This does not reject the thought that the Readme needs an overhaul.  In the end I wrote https://github.com/metamath/set.mm/wiki/Getting-started-with-contributing years ago, so I know there is headroom for improvement in general.

tirix:
I'd say let's be as complete as possible and add the instructions for Debian, marked as such. 
We can write something vague like "the process is similar for other distributions".
If someone else uses a different distro and can update the description for that case, it will be completed then.
Why be afraid of the worms in Pandora's box?...

wlammen:
For Debian users the easiest way to get metamath-exe is to follow one of the gcc commands listed in README.TXT.  The 'alternative approach' with autoreconf is for those having no gcc available or installed.  Then the configure script might provide a solution.

digama0:
I agree that we shouldn't let perfect be the enemy of the good. I am not suggesting that you or anyone else overhauls the README, although I wouldn't be opposed to that, only that we write down all the information that we have confirmed through user reports. No one has every OS and linux distro installed, so install instructions are necessarily crowd-sourced to some extent. Let's just put in all the information we learned from your experience, and if someone else has a bad experience on another distro we'll add that too, later.

tirix (approved):

wlammen:
@tirix Nice bonmot, by the way !

PR #188, opened 2026-07-13
Fix a few UBs

benjub:
This PR fixes a few UBs (undefined behaviors): a class of particularly critical bugs. In the present case, the fixed UBs were pointer arithmetic out of a valid object bounds, uninitialized variable, dangling pointer, out-of-bound read. These bugs could actually occur (and I actually got segfaults over the years because of the getFullArg one).

These UBs were found with TrustInSoft Analyzer and I was assisted in the bugfixes by claude code.

Maybe @tirix and/or @wlammen can have a look?

### DIFF
--- a/src/mmdata.c
+++ b/src/mmdata.c
@@ -1224,14 +1224,14 @@ void nmbrCpy(nmbrString *s, const nmbrString *t) {
 // Like strncpy, only the 1st n characters are copied.
 // Dangerous for general purpose use.
 void nmbrNCpy(nmbrString *s, const nmbrString *t, long n) {
-  long i;
-  i = 0;
-  while (t[i] != -1) { // End of string -- nmbrSeg, nmbrMid depend on it!!
-    if (i >= n) break;
+  long i = 0;
+  // Check i < n before reading t[i] so we never access t[n]; all callers
+  // set sout[n] = NULL_NMBRSTRING afterwards, so that value is never used.
+  while (i < n && t[i] != -1) { // End of string -- nmbrSeg, nmbrMid depend on it!!
     s[i] = t[i];
     i++;
   }
-  s[i] = t[i]; // End of string
+  s[i] = *NULL_NMBRSTRING; // End of string (NULL-terminate at actual end)
 }
 
 // Compare two strings.
--- a/src/mmdata.c
+++ b/src/mmdata.c
@@ -1225,8 +1225,9 @@ void nmbrCpy(nmbrString *s, const nmbrString *t) {
 // Dangerous for general purpose use.
 void nmbrNCpy(nmbrString *s, const nmbrString *t, long n) {
   long i = 0;
-  // Check i < n before reading t[i] so we never access t[n]; all callers
-  // set sout[n] = NULL_NMBRSTRING afterwards, so that value is never used.
+  // Only t[0..n-1] are guaranteed to be valid elements, so check i < n before
+  // reading t[i]. The result is terminated below, so callers do not have to
+  // terminate it.
   while (i < n && t[i] != -1) { // End of string -- nmbrSeg, nmbrMid depend on it!!
     s[i] = t[i];
     i++;
@@ -2706,9 +2707,9 @@ void pntrCpy(pntrString *s, const pntrString *t) {
 // Dangerous for general purpose use
 void pntrNCpy(pntrString *s, const pntrString *t, long n) {
   long i = 0;
-  // Check i < n before reading t[i] so we never access t[n] (which may be
-  // a dangling pointer when callers free the element at position n before
-  // calling this function).  All callers set sout[n] = NULL afterwards.
+  // Only t[0..n-1] are guaranteed to be valid elements, so check i < n
+  // before reading t[i]. The result is terminated below, so callers do not
+  // have to terminate it.
   while (i < n && t[i] != NULL) { // End of string -- pntrSeg, pntrMid depend on it!!
     s[i] = t[i];
     i++;
--- a/src/mmdata.c
+++ b/src/mmdata.c
@@ -2708,14 +2708,15 @@ void pntrCpy(pntrString *s, const pntrString *t) {
 // Like strncpy, only the 1st n characters are copied.
 // Dangerous for general purpose use
 void pntrNCpy(pntrString *s, const pntrString *t, long n) {
-  long i;
-  i = 0;
-  while (t[i] != NULL) { // End of string -- pntrSeg, pntrMid depend on it!!
-    if (i >= n) break;
+  long i = 0;
+  // Check i < n before reading t[i] so we never access t[n] (which may be
+  // a dangling pointer when callers free the element at position n before
+  // calling this function).  All callers set sout[n] = NULL afterwards.
+  while (i < n && t[i] != NULL) { // End of string -- pntrSeg, pntrMid depend on it!!
     s[i] = t[i];
     i++;
   }
-  s[i] = t[i]; // End of string
+  s[i] = *NULL_PNTRSTRING; // End of string (NULL-terminate at actual end)
 }
 
 // Compare two strings.
--- a/src/mmdata.c
+++ b/src/mmdata.c
@@ -1256,7 +1256,6 @@ temp_nmbrString *nmbrSeg(const nmbrString *sin, long start, long stop) {
   if (length < 0) length = 0;
   temp_nmbrString *sout = nmbrTempAlloc(length + 1);
   nmbrNCpy(sout, sin + start - 1, length);
-  sout[length] = *NULL_NMBRSTRING;
   return sout;
 }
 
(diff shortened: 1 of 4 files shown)


src/mmdata.c
```diff
@@ -2708,14 +2705,15 @@ void pntrCpy(pntrString *s, const pntrString *t) {
...
 // Dangerous for general purpose use
 void pntrNCpy(pntrString *s, const pntrString *t, long n) {
-  long i;
-  i = 0;
-  while (t[i] != NULL) { // End of string -- pntrSeg, pntrMid depend on it!!
-    if (i >= n) break;
+  long i = 0;
+  // Check i < n before reading t[i] so we never access t[n] (which may be
+  // a dangling pointer when callers free the element at position n before
+  // calling this function).  All callers set sout[n] = NULL afterwards.
```
david-a-wheeler: Comment is misleading, esp. last part. I suggest this instead:

```suggestion
  // Only t[0..n-1] are guaranteed to be valid elements, so check i < n
  // before reading t[i]. The result is terminated below, so callers of
  // pntrSeg/pntrMid/pntrLeft need not terminate it themselves.
```

src/mmdata.c
```diff
@@ -1224,14 +1224,14 @@ void nmbrCpy(nmbrString *s, const nmbrString *t) {
 // Like strncpy, only the 1st n characters are copied.
 // Dangerous for general purpose use.
 void nmbrNCpy(nmbrString *s, const nmbrString *t, long n) {
-  long i;
-  i = 0;
-  while (t[i] != -1) { // End of string -- nmbrSeg, nmbrMid depend on it!!
-    if (i >= n) break;
+  long i = 0;
+  // Check i < n before reading t[i] so we never access t[n]; all callers
+  // set sout[n] = NULL_NMBRSTRING afterwards, so that value is never used.
```
david-a-wheeler:
```suggestion
  // Only t[0..n-1] are guaranteed to be valid, so check i < n before
  // reading t[i]. The result is terminated below, so callers need not terminate it.
```

src/mmpars.c
david-a-wheeler: This doesn't make sense to me. Just say "} else {" without "if (1)", and delete the line saying "bug(1746)". If you're going to use if (1), then the bug(1746) cannot *ever* trigger.

src/mmpars.c
```diff
@@ -5005,7 +5005,7 @@ void getNextInclusion(char *fileBuf, long startOffset, // inputs
         }
         continue; // We're past Begin; start search for End
       } // Begin, End, or Skip
-    } else if (i != i + 1) { // Suppress "unreachable code" warning for bug trap below
+    } else if (1) { // Suppress "unreachable code" warning for bug trap below
```
david-a-wheeler:
```suggestion
    } else {
```

src/mmpars.c
david-a-wheeler:
```suggestion

```

david-a-wheeler:
Overall looks good, definite improvement, but see my comments above. Once fixed, let's merge.

src/mmdata.c
benjub: Counter-suggestion, parallel to your other suggestion:
```suggestion
  // Only t[0..n-1] are guaranteed to be valid elements, so check i < n
  // before reading t[i]. The result is terminated below, so callers do not
  // have to terminate it.
```

src/mmpars.c
benjub: I almost did that and deleted "bug(1746)", but then I wondered if "bug(1746)" was a sentinel there to make sure we never went there, and the "if (1)" was so that compilers would not optimize away and "delete" bug(1746), defeating its purpose. But that woud be very weird indeed. That was probably a remnant of older code.

Given your comment, then indeed, we can have a simple "else" and delete "bug(1746)".

src/mmdata.c
benjub:
```suggestion
  // Only t[0..n-1] are guaranteed to be valid elements, so check i < n before
  // reading t[i]. The result is terminated below, so callers do not have to
  // terminate it.
```

benjub:
> Overall looks good, definite improvement, but see my comments above. Once fixed, let's merge.

Thanks for the review (my comments were suboptimal indeed). I applied your suggestions (with slight comment edits). Feel free to modify them if necessary and merge.

david-a-wheeler (approved):

david-a-wheeler:
Thanks for the fixes!

benjub:
> Thanks for the fixes!

Thanks for the review!

david-a-wheeler:
I don't like having UBs in metamath-exe. I created a trivial fuzzer & found several. So stay tuned, I'm going to soon create a branch with a trivial fuzzer & fixes for the UBs it found. Once we get those under control, using LTO will be far less scary.

wlammen:
I recently started to work a bit on the Metamath program again, actually prompted by this pull request, and the fact that my name was mentioned in it.

I found several issues concerning the C sources, but instead of fixing them randomly in an uncoordinated fashion, I prefer working bottom up: Get the fundamentals right first, before addressing issues on higher levels.

On the lowest level is documentation.  The build.sh script with options -cd allows you to automatically generate doxygen documentation, if only you have doxygen installed.  Documentation is possible without changing Metamath's behavior, and hence needs little more than a review cycle.

The next level of infrastructure is a regression test framework that should be available on a broader range. This would permit changes to the source code without introducing hidden bugs  -  well, at least, reduce that risk, and keeping interfaces stable while working under the hood.

Then one could advance the currently supported C standard, which is C99. Really. 25 years old. I recommend to allow C17 instead, still a decade old, but C23 is perhaps too fresh to require its use.

Once the technical details are agreed upon, I'd start writing a reference counting system. This emulates the OOP style found in more modern languages, and could in principle solve resource allocation conflicts, and automatic resource handling.

One of the first candidates is the string representation in mmvstr.*.  A had a closer look at it, and, my God, found plenty of issues with its implementation.  The idea behind this framework is an implementation of a BASIC language version.  Nobody knows anything about this language any more, and the interface looks outdated. So while reworking it, some Unicode compatibility could also be addressed.

There is more to do, but perhaps this is enough for a first round.

The first results are available on my computer, a documentation of the test framework.

Are you interested? And how should the results be sent? Piece for piece in pull request, or sending in complete source files?

sctfn:
Personally I’d love Unicode support. metamath-exe is the only validator that doesn’t support Unicode. If we get that, we can update the spec and start using Unicode to set.mm. On Aug 4, 2026, at 8:27 AM, Wolf Lammen ***@***.***> wrote:﻿wlammen left a comment (metamath/metamath-exe#188)
I recently started to work a bit on the Metamath program again, actually prompted by this pull request, and the fact that my name was mentioned in it.
I found several issues concerning the C sources, but instead of fixing them randomly in an uncoordinated fashion, I prefer working bottom up: Get the fundamentals right first, before addressing issues on higher levels.
On the lowest level is documentation.  The build.sh script with options -cd allows you to automatically generate doxygen documentation, if only you have doxygen installed.  Documentation is possible without changing Metamath's behavior, and hence needs little more than a review cycle.
The next level of infrastructure is a regression test framework that should be available on a broader range. This would permit changes to the source code without introducing hidden bugs  -  well, at least, reduce that risk, and keeping interfaces stable while working under the hood.
Then one could advance the currently supported C standard, which is C99. Really. 25 years old. I recommend to allow C17 instead, still a decade old, but C23 is perhaps too fresh to require its use.
Once the technical details are agreed upon, I'd start writing a reference counting system. This emulates the OOP style found in more modern languages, and could in principle solve resource allocation conflicts, and automatic resource handling.
One of the first candidates is the string representation in mmvstr.*.  A had a closer look at it, and, my God, found plenty of issues with its implementation.  The idea behind this framework is an implementation of a BASIC language version.  Nobody knows anything about this language any more, and the interface looks outdated. So while reworking it, some Unicode compatibility could also be addressed.
There is more to do, but perhaps this is enough for a first round.
The first results are available on my computer, a documentation of the test framework.
Are you interested? And how should the results be sent? Piece for piece in pull request, or sending in complete source files?

wlammen:
Unicode support: Maybe not full support in the first step.  I hate to think about mixing left-to-right and right-to-left scripts in a text string, and what to do, if texts need to be truncated because of buffer limitations.  A simple form of UTF-8 support should be possible, though.  Actually I already spotted a Unicode related issue in mmfatl.c: Generation of fatal error messages use the `__FILE__` macro in reporting error locations. This macro consists of a file name (all ASCII for Metamath) and their directory path not necessarily ASCII on personal computers. I fixed this on my computer already.

david-a-wheeler:
@wlammen - I suggest creating a discussion in a new issue. This is an already-merged PR, and it's a miracle I noticed your comment :-).

Updating to a newer version of C sounds exactly correct. Yes yes! That's also something that can be done incrementally, greatly reducing the likelihood of a subtle problem.

Adding UTF-8 also seems correct. It might not even be that hard.

I'm way more skeptical of adding reference counting. That's obviously possible, but there are also many reasons to avoid it.

There are a lot of "general cleanups" that could be done incrementally. I'd start there.

PR #189, opened 2026-07-21
Portable build

david-a-wheeler:


david-a-wheeler:
This collection of commits is intended to make it *MUCH* easier for people to use metamath-exe, without ever needing to compile or even necessarily install:

* It creates a single cosmopolitan "Actually Portable Executable" (APE) executable file that runs on Windows, MacOS, Linux, and BSD*, on Intel or ARM.
* It creates a WASM build and HTML page so you can run Metamath directly from your web browser (no install required at all).

david-a-wheeler:
@benjub @digama0 - thoughts?

david-a-wheeler:
The cosmopolitan build is 2.7 MiB. That's bigger than any one build, but it's not too bad, and it's much smaller than set.mm.

Here's what it looks like on a web browser:

image

README.TXT
```diff
@@ -17,6 +17,99 @@ information that applies to the content of this package.
...
+            "metamath-vX.Y.Z.exe".  Because the file is not code-signed, you
+            may see a Microsoft Defender SmartScreen "Windows protected your
+            PC" prompt; click "More info" then "Run anyway".  (Downloading the
+            version-numbered filename, rather than a plain "metamath.exe",
+            reduces the chance of a false-positive antivirus warning.)
+
+  macOS:    the file is not notarized, so Gatekeeper quarantines downloads.
+            Clear the quarantine flag once, then run it from Terminal:
+
+              xattr -d com.apple.quarantine metamath-vX.Y.Z.exe
```
benjub, line 63: Above, you write "Each release also contains...", so it means that `metamath.exe` also works, correct? That would probably make for a simpler command to type, here and below (but if you choose to keep the name with the version, I think there is should not be a `.Z`, given the current version).

README.TXT
```diff
@@ -17,6 +17,99 @@ information that applies to the content of this package.
...
+way to save files the program produced back to your computer.  A file added
+this way stays in your browser; it is not sent anywhere.
+
+Note that the databases live only in the virtual filesystem, which is this
+page's own storage.  Downloading set.mm does not read it into metamath; you
+still have to type
+
+  read "set.mm"
+
+To build it yourself you need the Emscripten SDK.  From the top folder:
```
benjub: To build **what** yourself?  Maybe write: "To build this browser version locally, you need...".

benjub:
> @benjub @digama0 - thoughts?

Anything that makes the program more accessible is a good idea in my opinion.  I didn't know about that kind of "ultimate portability"!  Have you tested the performance of that universal portable against a metamath built from source (e.g., how long to verify set.mm when running both on a same controlled linux system)?  If the difference is notable, that should be mentioned.

By the way: since it will make metamath.exe more accessible and hopefully more used, it's even more important to fix its UBs, like in #188 ;)

README.TXT
david-a-wheeler: Just "metamath" would work. The ".Z" is just an example to make it clear it's a version# and not a floating point number.

My plan is to distribute *with* and *without* the version#. The version number is important for Windows; Windows defender gets cranky if there's no version number.

README.TXT
david-a-wheeler: Sorry, yes, that's correct. The *browser* build needs Emscripten. You of course don't need Emscripten if you want to use a normal C compiler and compile as usual.

README.TXT
david-a-wheeler: We can pick the names, of course. The *easiest* would be to always have the same fixed name no matter what, though Windows might be slightly unhappy sometimes.

david-a-wheeler:
> Anything that makes the program more accessible is a good idea in my opinion.

Thanks! I think so too!

> I didn't know about that kind of "ultimate portability"! Have you tested the performance of that universal portable against a metamath built from source (e.g., how long to verify set.mm when running both on a same controlled linux system)? If the difference is notable, that should be mentioned.

I have not, but I expect *no* performance loss from the cosmopolitan APE build compared to typical compiles, because metamath is generally compute-bound (not I/O-bound). If anything, it'll be a *gain* compared to how many people compile metamath-exe.

Cosmopolitan takes a very simple approach to compiling. Since Cosmopolitan 3.x, cosmocc builds a fat APE: it compiles the code twice (once for x86-64, once for aarch64) and bundles both native architectures into the single executable. The APE loader picks the matching native slice at launch, and interposes various calls that end up interacting with the OS kernel. So there's a *very* slight I/O overhead, but no compute overhead.

If anything, the performance is better. Many people just do a simple `gcc -o metamath *.c` which *works* but doesn't ask for optimization. If you use the `autoconf` version, that takes longer to build, but it enables `-O3` optimization which produces a *much* faster executable. I built this cosmopolitan approach via autoconf, so it inherits all the `-O3` goodness.

Nothing is free, of course. The resulting file is much bigger. It includes *two* different CPU architectures (and you only need one), and it includes various header shims to work on various operating system interfaces (which you don't need for any one system). That means it takes more time to download and more storage space. However, I think the *ease* of "download and run" is worth it for most, and for most, the storage space & network time is not important at these scales.

That is *NOT* true for the WASM (web browser) build using Emscripten, which *will* be slower than a "real" executable. Verifying *all* the proofs takes ~12 seconds on my system for the browser version, vs. ~6 seconds when compiled normally. So it's definitely *slower* (about 1/2 speed) when run from a web browser, but it's *usable*.

> By the way: since it will make metamath.exe more accessible and hopefully more used, it's even more important to fix its UBs, like in https://github.com/metamath/metamath-exe/pull/188 ;)

I agree completely! We should fix its UBs, no doubt.

In fact, there's another incentive as well: I turned on `-O3` optimization, but I intentionally did *not* turn on link time optimization (lto). This option does the moral equivalent of putting all the source files into 1 file and compiling it. This enables a number of other optimizations, which *might* make the result faster, but lto is really good at unintentionally triggering UBs, so I intentionally didn't add it. If we had more confidence that we'd fixed UBs, we could turn on that optimization and make it run faster.

david-a-wheeler:
A few corrections: the expectation is that .exe is part of the name, since Windows in particular expects it, while other operating systems *can* use it that way.

README.TXT
david-a-wheeler: I modified the text to *hopefully* make that clearer.

david-a-wheeler:
I tried to improve the Emscripten speed (which generates the web browser version) using `-O3`, as that's the setting already used by Cosmopolitan. It didn't really help, at least in my initial tests. I also documented, in comments, likely ways to improve the performance of the web browser distribution (using wasm) and why I didn't immediately do them. I think we could get somewhat better performance with more effort, but step 1 is getting a working one released at all :-).

The `build-wasm.sh` script documents how we might improve the wasm build's performance further:

* -flto    Link-time optimization would let the optimizer see across all
  translation units.  We avoid it: the code has some undefined
  behavior, and LTO's whole-program view lets the compiler exploit
  UB more aggressively (and drop code it "proves" unreachable),
  which risks miscompiles.  Revisit once the UB is cleaned up.
*  -sASYNCIFY_ONLY / -sJSPI
  Plain -sASYNCIFY (used below) instruments EVERY function that
  could be on the stack at a yield, which slows all hot code
  (e.g. the proof verifier), not just the I/O that actually yields.
  -sASYNCIFY_ONLY=[...] restricts instrumentation to the functions
  on the yield path (cmdInput()/print2() and their callers), which
  reclaims most of the overhead -- but it requires hand-maintaining
  that whitelist, and omitting a function is a runtime error.
  -sJSPI replaces ASYNCIFY with the browser's native stack
  switching (no instrumentation overhead) but is not yet supported
  across all browsers.  Both are deferred as additional work.

david-a-wheeler:
>  I didn't know about that kind of "ultimate portability"!

It's certainly not known everywhere! It's an interesting bag of tricks, but when the result makes users' lives easier, I think it's worth exploring. This was my first time using the cosmo tools, and they were remarkably easy to use for this case (no GUI).

benjub:
> > By the way: since it will make metamath.exe more accessible and hopefully more used, it's even more important to fix its UBs, like in #188 ;)
> 
> I agree completely! We should fix its UBs, no doubt.

Well, you have merge rights!

david-a-wheeler:
Thanks! I'm going to make some improvements to the web interface. It *works* now, but there's no "real OS" behind it, so it's awkward to use in practice (and can easily lose data since it's all *only* in memory). I'm going to tweak the web interface version so it'll be easier to use & won't easily throw away changes.

benjub:
> Thanks! I'm going to make some improvements to the web interface. It _works_ now, but there's no "real OS" behind it, so it's awkward to use in practice (and can easily lose data since it's all _only_ in memory). I'm going to tweak the web interface version so it'll be easier to use & won't easily throw away changes.

It may have been lost in the discussion thread, but what I meant is: you have the merge rights for #188, which *does* fix a few UBs, which you agreed ought to be fixed!

david-a-wheeler:
@benjub - Thanks. I *did* understand, but my brain works better if I focus one-at-a-time. I'll definitely look at #188, though, it's *clearly* important that we fix UBs.

Minor annoyance: I *really* want to prevent data loss in the web interface. It's a simulated filesystem, so you'd think I should use the simulated filesystem interface for JavaScript called "OPFS". It turns out that's complicated, not as well supported, and completely unnecessary. OPFS supports in-the-middle operations we don't need, we just need a safe storage system. It turns out the older IDBFS works fine for our purposes and works everywhere... and working reliably is kind of the point. So, out with the old, in with the new. I think it's unfair to give people a web interface and make it easy to lose data, that would make people sad. I don't want sad people :-).

david-a-wheeler:
@benjub -  as you know, I separately merged commits that greatly reduce UBs in metamath-exe in general, which I was concerned about before merging this one.

Any last-minute objections to merging this PR?

src/mminou.c
```diff
@@ -81,7 +81,50 @@ long g_screenHeight = SCREEN_HEIGHT; // Default = 23
...
+ * VERIFY PROOF * would hold that thread for many seconds.  Nothing already
+ * written could be painted, and the page would appear to have frozen.
+ * Yielding briefly lets the browser draw the output produced so far and stay
+ * responsive.  Yields happen no more than every 50 milliseconds, so the cost
+ * is negligible compared with the work being done.
+ */
+/*! Milliseconds between yields.  Larger is faster but less smooth. */
+#ifndef MM_YIELD_MS
+# define MM_YIELD_MS 50
+#endif
```
benjub: Move above the previous doxygen comment, else comments won't be attached to the correct macro/function.

src/metamath.c
benjub, line 1190:
```suggestion
#ifdef __EMSCRIPTEN__
        // The page supplies its own scrollback, and the prompted pager waits
        // in getchar() for an answer the browser has no way to deliver.
        print2("?Prompted scrolling is not available in the browser.\n");
#else
        defaultScrollMode = 1;
        g_scrollMode = 1;
        print2("Prompted scrolling is now in effect.\n");
#endif
```
The comment itself was suggested by AI. I think that this is the only remaining place where we need to guard against scrolling.

benjub:
Two (final?) comments above.

I have looked only at the src/ changes, since I do not know enough to review the rest. So probably safer, after the last edits, to rebase and retest the wasm build.

src/mminou.c
david-a-wheeler: I agree! I'll submit a commit to fix that.

src/metamath.c
david-a-wheeler: I agree. I'll submit a commit to fix that. I'll also ensure that CONTINUOUS is the default when EMSCRIPTEN is enabled.

david-a-wheeler:
A rebase would cause the past conversation hashes to change & overall be weird, and there's a lot of conversation history here. I'll install merge `master` into this branch. git is quite capable of handling that (it's a normal operation).

david-a-wheeler:
I tested the wasm build, all works. However, the cosmopolitan build and "normal" build share a directory by default for their intermediates, and this can be a footgun for testing. I'll add a commit to use a different directory for cosmopolitan, for people who build & test both. That won't impact normal *users* of metamath-exe, but it can impact people who modify its code :-).

david-a-wheeler:
I'm running tests in my own directory for many combos. I expect that to pass. If so, I think we're ready to merge, and I think we should do that soon so we can use it.

david-a-wheeler:
Eventually the CI should build the wasm build, but there are so many commits here I want to *STOP*, merge, and then move on in a reasonable way.

benjub:
> Eventually the CI should build the wasm build, but there are so many commits here I want to _STOP_, merge, and then move on in a reasonable way.

I agree: merge it at your earliest convenience. Then I think there is a kind of urgent matter needing your attention at https://github.com/metamath/metamath-website-scripts/issues/21#issuecomment-5260998228

And then, if you can approve and/or merge #194 and #195, that'd be great. They are ready to be merged according to me (barring some merge conflicts to resolve once this one is merged).

david-a-wheeler:
> And then, if you can approve and/or merge https://github.com/metamath/metamath-exe/pull/194 and https://github.com/metamath/metamath-exe/pull/195, that'd be great. They are ready to be merged according to me (barring some merge conflicts to resolve once this one is merged).

Sorry, I *just* opened them to review (I've been busy on other things), and see that they're already merged. I think that's fine. If we find further problems, we'll fix those. I'm hoping that my newly-added tests, and more tests to come, will reduce risks of further problems.

david-a-wheeler:
> I think there is a kind of urgent matter needing your attention at https://github.com/metamath/metamath-website-scripts/issues/21#issuecomment-5260998228

Yes, that was WAY more important, so I focused on that first :-).

> I agree: merge it at your earliest convenience. 

Great! I've re-merged the updated master into this, as a double-check. Everything still works, so I'm merging.

PR #190, opened 2026-07-28
Fix more ubs

david-a-wheeler:


david-a-wheeler:
@benjub - this was more than I expected.

I want to eliminate obvious undefined behaviors (UBs) so that we can use LTO optimization. I implemented a simple fuzzer that enabled some sanitizers, expecting to find a few things, fix them, and call it a day.

The simpler fuzzer turned out to find a *lot* of problems. Practically all were trivial (e.g., a lot of bounds-checking before use), but there sure were a number. As I fixed them *eventually* the find rate went down (as expected, but that took longer than expected).

I think this is a good thing. Comments welcome.

david-a-wheeler:
Sorry it's so long, but each commit is relatively simple.

benjub:
> @benjub - this was more than I expected.
> 
> I want to eliminate obvious undefined behaviors (UBs) so that we can use LTO optimization. I implemented a simple fuzzer that enabled some sanitizers, expecting to find a few things, fix them, and call it a day.
> 
> The simpler fuzzer turned out to find a _lot_ of problems. Practically all were trivial (e.g., a lot of bounds-checking before use), but there sure were a number. As I fixed them _eventually_ the find rate went down (as expected, but that took longer than expected).
> 
> I think this is a good thing. Comments welcome.

That's impressive, and very good improvements to have eliminated all these UBs. It's nice that the test suite got expanded in the process too. Looking at the commit titles (especially the later commits), I can't be sure: are they all only about UB fixing, or are there also functional changes?

What are your plans relative to the limitations you wrote, and concerning afl++ and libFuzzer?  If you plan to do more UB-hunting in the not-too-distant future, I would prefer to wait before doing another UB-hunting pass/verification with TrustInSoft Analyzer.

david-a-wheeler:
Thanks!

> That's impressive, and very good improvements to have eliminated all these UBs. It's nice that the test suite got expanded in the process too. Looking at the commit titles (especially the later commits), I can't be sure: are they all only about UB fixing, or are there also functional changes?

Thanks! There is *no* functional change for correct inputs in this branch.

My focus in this branch is *solely* about fixing undefined behavior (UB). It's too dangerous to enable LTO if the code is riddled with UBs. It also irks me that there *are* UBs. This program handles untrusted data from others, and its purpose is to rigorously verify correctness. In my mind, it shouldn't have *any* UBs. If we take steps to find and eliminate UBs, the result is better for all. While other tools have advantages, fixing the UBs in this tool isn't *that* hard and "should be done".

In *many* cases, the fix was an "add a range check before using an index value" kind of fix.

However, in a few cases it wasn't *quite* that simple. I tried to make each change as "clearly correct" as I could.

> What are your plans relative to the limitations you wrote, and concerning afl++ and libFuzzer? If you plan to do more UB-hunting in the not-too-distant future, I would prefer to wait before doing another UB-hunting pass/verification with TrustInSoft Analyzer.

I intend to keep this fuzzer in the repo, it's useful. However, fuzzers generally find lots of problems at first, and as those get fixed, they find fewer and fewer. I ran the fuzzer for *many* hours today, and it only found one more. I fixed that one. I plan to run it overnight, but I don't expect it to find much (maybe 1, but 0 would also be unsurprising).

So I *think* I'm at the end for this branch.

My request to you:

1. Could you or someone else take a look and approve (or let me know if you see any issues)? There are a lot of changes, but except for the fuzzer (which is an internal development tool) they're relatively small & bounded to eliminating UBs (there's no functional change).
2. I currently do *NOT* plan to use afl++ or libfuzzer on metamath-exe. If someone else wants to.. great! I'm a fan of afl++.
3. Once this branch is merged (as a merge commit), please *do* try out another tool to find UBs. Different tools tend to find different problems. But let's not waste tools on problems fixed in this branch. Let's merge these fixes, and then other tools can try to focus on what's left.

david-a-wheeler:
All of these fixes involve fixing a UB, since the fuzzer can only detect UBs. However, where possible, I focused on fixing root causes. For example, in commit e9e08b9984e3f6170c146f0d6f7717bf62631bdd the UB was because "$$" was parsed differently by different parts of the program. Since "$$" isn't in the Metamath spec and it isn't used by databases, removing its partial-and-wrong support eliminated the UB.

david-a-wheeler:
@benjub - I had *no* idea how many stupid little bugs I'd find. I believe *none* of these impact correct databases and correct queries, but there were a bunch of triggerable UBs that this branch *fixes*.

I know there's a lot of little changes, but could you take at least a glance & tell me if these make sense? I'd like to get all these fixes merged, so they trouble us no more.

src/mmvstr.c
```diff
@@ -390,7 +425,19 @@ temp_vstring edit(const char *sin, long control) {
 
   // Copy string
   i = (long)strlen(sin) + 1;
-  if (untab_flag) i = i * 7; // Allow for max possible length
+  if (untab_flag || tab_flag) {
+    // Allow for max possible length.  The tab-expansion loop further below
+    // runs for either flag, not just for untab_flag, and it replaces each
+    // tab with as many as 8 spaces.  That loop's length counter also ends
+    // up one position further ahead per tab, so allow 8 extra characters
+    // per tab (7 for the expansion, 1 for the counter) plus a little slack.
```
benjub: Rather than allocating extra bytes to avoid out-of-bound, it'd be better to fix the off-by-one counting L613.

src/mmvstr.c
benjub:
```suggestion
      k = k + 7 - ((m - 1) & 7);
```

src/mmvstr.c
```diff
@@ -808,7 +857,15 @@ temp_vstring str(double f) {
...
-  sprintf(s,"%f", f);
+  // Bound the write by the size just allocated.  "%f" prints every digit
+  // before the point, so a large enough f needs far more than 50 bytes:
+  // 1e300 expands to over 300 characters.  No caller passes anything like
+  // that today -- the largest is a long, which fits in 26 -- so this only
+  // changes what happens if one ever does, from overrunning the buffer to
+  // truncating.  It also silences a bogus "null destination pointer"
+  // warning from the fortified sprintf() under -fsanitize; that warning
+  // fires even for a plain local array, so it is not about s being null.
+  snprintf(s, 50, "%f", f);
```
benjub: That 50 should programmatically stay equal to the 50 in `temp_vstring s = tempAlloc(50);`.
So maybe use `const long size = 50;`

src/mmvstr.c
benjub:
```suggestion
        while (j - 1 > i - 8 && sout[j - 1] == ' ') {
```
No UB here, but for consistency with outer loop.

src/mmcmds.c
```diff
@@ -4528,9 +4531,13 @@ void eraseSource(void) // ERASE command
...
   // g_MathToken[g_mathTokens].tokenName is assigned in
-  // parseMathDecl() by let().  eraseSource() should free every g_MathToken and
-  // there are (g_mathTokens + g_dummyVars) tokens.
-  for (i = 0; i <= g_mathTokens + g_dummyVars; i++) {
+  // parseMathDecl() by let().  eraseSource() should free every g_MathToken.
+  // The entries run contiguously: the declared symbols, the "$|$" boundary
+  // token, any placeholders the parser made for undeclared symbols, a second
+  // boundary token if there were placeholders, then the dummy variables.
+  // HIGHEST_MATH_TOKEN is that highest entry.  (Using g_mathTokens here
+  // instead used to leave any placeholder above g_dummyVars unfreed.)
```
benjub:
```suggestion
  // instead would leave any placeholder above g_dummyVars unfreed.)
```
Do not comment older versions of the code.

src/mmvstr.c
```diff
@@ -198,10 +209,20 @@ void let(vstring *target, const char *source) {
...
   if (sourceLength) {
-    strcpy(*target, source);
+    // A few callers deliberately write let(&x, x), only to trigger the
+    // freeTempAlloc() below.  strcpy() is undefined when its arguments
+    // overlap (both are restrict-qualified), and here there is nothing to
+    // copy anyway, so skip just the copy.
+    // This test has to stay nested inside "if (sourceLength)": folding it
+    // into that condition would send a self-assignment to the else branch,
+    // which would wrongly replace the string with "".
+    if (*target != source) strcpy(*target, source);
```
benjub: Why not use memmove?
```suggestion
    // memmove() is defined for overlapping arguments, unlike strcpy() whose
    // parameters are restrict-qualified.  That covers the callers who write
    // let(&x, x) purely to trigger the freeTempAlloc() below (mmunif.c and
    // mmwtex.c), and any future caller passing a pointer into *target.
    // The length is already known, so this is no slower than strcpy().
    // This memmove() has to be guarded by "if (sourceLength)": a non-empty
    // self-assignment in the else branch would wrongly free the string and
    // replace it with "".
    memmove(*target, source, sourceLength + 1);
```
(comment mostly generated by AI)

src/mmvstr.c
```diff
@@ -109,18 +109,23 @@ void freeTempAlloc(void) {
...
+ *   [noreturn] If full, the program terminates with error code EXIT_FAILURE.
  */
 static void pushTempAlloc(void *mem)
 {
   if (g_tempAllocStackTop >= (MAX_ALLOC_STACK-1)) {
+    // Report this one directly rather than through bug().  bug() builds
+    // its messages with let() and print2(), which allocate temporaries of
+    // their own, so with this stack already full it re-enters
+    // pushTempAlloc(), reports the same overflow again, and recurses
+    // until the C stack runs out.
```
benjub, line 124: Maybe also write a similar comment at the definition of bug()? Claude proposed to write there:
```
  /* pushTempAlloc() in mmvstr.c prints this same line directly rather than
     calling bug(), because bug() allocates temporaries and would recurse
     when the temporary-string stack is what overflowed.  Keep them in sync. */
  print2("?BUG CHECK:  *** DETECTED BUG %ld\n", (long)bugNum);
```

After discussion, Claude proposed a macro: in mmdata.h, write
```
/*! Format of the line bug() prints when it detects a bug.  pushTempAlloc()
    in mmvstr.c prints this itself instead of calling bug() -- see the
    comment there -- so keep the two in sync.  */
#define BUG_CHECK_FORMAT "?BUG CHECK:  *** DETECTED BUG %ld\n"
```
and then use that macro in both mmdata.c:747:
```
  print2(BUG_CHECK_FORMAT, (long)bugNum);
```
and mmvstr.c:124:
```
    printf(BUG_CHECK_FORMAT, 2201L);
```
although that introduces a dependence of mmvstr.c on mmdata.h.

src/mmdata.h
benjub, line 260: That "statement_struct stays 256 bytes" as you write should be programmatically checked if possible. The size depends on the architecture, and I'm not sure how to do it, but claude told me the following would work. I made it converge to a simpler solution, but please check:
```
/* Keep statement_struct at four 64-byte cache lines.  hasVarWithoutHyp is
   placed in the padding before "scope" for that reason; growing the struct
   to 264 slowed "verify proof *" over set.mm from 7.84 s to 9.41 s.  */
#if defined(__LP64__) || defined(_LP64)
_Static_assert(sizeof(struct statement_struct) == 256,
    "statement_struct grew -- see the hasVarWithoutHyp comment above");
#endif
```

src/mmpars.c
```diff
@@ -318,6 +327,13 @@ void parseKeywords(void)
         }
         // Initialize a new statement
         g_statements++;
+        // g_Statement[] was sized from readRawSource()'s count, and a dummy
+        // statement is written at g_statements + 1 after this loop, so both
+        // indexes have to stay inside the array.  This should always hold:
+        // this function creates a statement only where readRawSource()
+        // counted one.  Trap it rather than corrupt the heap if the two ever
+        // drift apart again, the way they did over "$$".
+        if (g_statements + 1 >= potentialStatements) bug(1774);
```
benjub: Isn't it similar to your other fixes ? bug() returns, as you noticed.
```suggestion
        if (g_statements + 1 >= potentialStatements) {
          bug(1774);
          exit(EXIT_FAILURE);  // bug() returns on "I"; see the comment at bug(2202)
        }
```

src/mmpars.c
```diff
@@ -3436,10 +3676,13 @@ void rawSourceError(char *startFile, char *ptr, long tokLen, vstring errMsg) {
...
   }
-  endLine--;
-  let(&errLine, space(endLine - startLine + 1));
-  if (endLine - startLine + 1 < 0) bug(1721);
-  memcpy(errLine, startLine, (size_t)(endLine - startLine) + 1);
+  // endLine now points just past the last character of the line, so the line
+  // length is endLine - startLine.  Do not decrement endLine to make it point
+  // at the last character: for a 0-length line that would compute a pointer
+  // before the start of the buffer, which is undefined behavior.
+  if (endLine - startLine < 0) bug(1721);
```
benjub: Claude tells me the problem is the same here, although currently not triggered. For future-proofing, I would add something like:
```suggestion
  if (endLine - startLine < 0) {
      bug(1721);
      exit(EXIT_FAILURE);  // bug() returns on "I"; see the comment at bug(2202)
  }
```

benjub (commented):
First part of my AI-assisted review. (Posting it now since github looks shaky and I'm afraid to lose the comments).

src/mmpars.c
```diff
@@ -2007,9 +2186,21 @@ char parseProof(long statemNum)
   // fbPtr points to the first token now.
 
   // First break up proof section of source into tokens
+  //
+  // Stop at the recorded end of the proof section rather than relying on
+  // proofTokenLen() to stop at the "$." that ends it.  It runs through a "$"
+  // followed by a digit, that being the "$1" dummy variable form, and through
+  // a "$$" as well, so a token can swallow the "$." and the scan carry on into
```
benjub: Isn't it precisely what you fixed, so not the case anymore?

src/mmpars.c
```diff
@@ -2908,9 +3110,22 @@ char parseCompressedProof(long statemNum)
   // ****** in order to easily parse the label section.
 
   // First break up the label section of proof into tokens
+  //
+  // Bounded by the recorded end of the proof section for the same reason as
+  // the scan in parseProof(): proofTokenLen() runs through "$$" and through a
+  // "$" followed by a digit, so a token can swallow the "$." that ends the
```
benjub: Isn't it precisely what you fixed, so not the case anymore?

src/mmpars.c
```diff
@@ -551,12 +578,58 @@ void parseLabels(void) {
   }
 }
 
+/*!
+ * \brief tokenLen() bounded by the end of the section being scanned
+ *
+ * The token length functions cannot be relied on to stop where a statement
+ * does.  tokenLen() runs through a "$" followed by a digit, that being the
+ * "$1" dummy variable form, so a token can reach past the "$." that ends the
```
benjub: Is that still correct after your fixes?

src/mmpars.c
```diff
@@ -3751,10 +4013,7 @@ long tokenLen(char *ptr)
   while (1) {
     tmpchr = ptr[i];
     if (tmpchr == '$') {
-      if (ptr[i + 1] == '$') { // '$$' character
-        i = i + 2;
-        continue;
-      } else {
+      {
```
benjub: You could actually remove that brace and its closing one, and de-indent the block.

src/mmpfas.c
```diff
@@ -2828,7 +2884,7 @@ void replaceDummyVar(long dummyVar, const nmbrString *mString)
     nmbrTmpPtr = (g_ProofInProgress.target)[step];
     slen = nmbrLen(nmbrTmpPtr);
     for (sym = slen - 1; sym >= 0; sym--) {
-      if (nmbrTmpPtr[sym] == dummyVar + g_mathTokens) {
+      if (nmbrTmpPtr[sym] == dummyVar + g_dummyVarBase) {
```
benjub, line 2887: Can you check if similar substitutions should be done on/around lines 839, 911, 1103, 2100, 2616, 2975, 3002, 3036 ? In ambiguous cases, a comment in the code may help.

src/mmpfas.h
```diff
@@ -18,6 +18,21 @@ extern flag g_proofChangedFlag; /*!< Flag to push 'undo' stack */
...
+ * token, any placeholders the parser made for undeclared symbols, a second
+ * boundary token if there were placeholders, then the dummy variables.
+ * \ref g_dummyVarBase indexes that last boundary token, so the dummy
+ * variables occupy g_dummyVarBase + 1 through g_dummyVarBase +
+ * \ref g_dummyVars.
+ *
+ * Both the place that grows the array and the place that frees it need this
+ * number; computing it in one place keeps them from disagreeing.
+ */
+#define HIGHEST_MATH_TOKEN (g_dummyVarBase + g_dummyVars)
```
benjub: Maybe declare it in mmdata.h ?

src/mminou.c
```diff
@@ -1283,6 +1287,44 @@ vstring fGetTmpName(const char *filePrefix) {
...
+ * guarantees a match at or before the end, so the one outcome strchr() cannot
+ * describe -- reaching the end with no match -- cannot happen.
+ *
+ * A NUL of s's own also stops strchr(), and it reports that as NULL.  The
+ * position is wanted in that case too, so fall back to strlen() for it.  That
+ * is a second scan, but only for a file that is about to be rejected.
+ *
+ * \param[in,out] s the string to scan.  s[len] is written and restored, so it
+ *   must be writable; it is the terminating NUL the caller just wrote.
+ * \param[in] len the offset of s's terminating NUL
```
benjub:
```suggestion
 * \param[in,out] s the string to scan.  Must be writable at s[len]: the
 *   sentinel is written there and restored before this returns.
 * \param[in] len the offset of s's terminating NUL.  This is a requirement,
 *   not a description: s[len] must be that NUL.  Pass anything else and the
 *   sentinel is written outside the string, the wrong byte is put back in its
 *   place, and nothing reports it.
```

src/mminou.c
```diff
@@ -1390,15 +1434,42 @@ vstring readFileToString(const char *fileName, char verbose, long *charCount) {
...
+  // already refuses a file for the same reason.
+  //
+  // Both answers come out of the one scan this function already made.
+  // scanToCharOrEnd() is the scan for carriage-returns that was here before,
+  // keeping the position it used to discard, and that position is what says
+  // whether the scan ended at the terminator or at a NUL of the file's own.
+  // When it ends at a carriage-return instead, the clean-up loop below walks
+  // every remaining byte anyway, so it carries the rest of the check.
+  nulPos = -1; // Offset of a NUL belonging to the file; -1 if there is none
+  scanStop = scanToCharOrEnd(fileBuf, *charCount, '\r');
```
benjub, line 1463:
```suggestion
  // fileBuf[*charCount] is the terminating NUL written above -- and rewritten
  // by the Unicode branch, which resets *charCount with it -- which is what
  // scanToCharOrEnd() requires.
  scanStop = scanToCharOrEnd(fileBuf, *charCount, '\r');
```

fuzz/run-cases.sh
```diff
@@ -0,0 +1,73 @@
...
+unexpected=0
+
+for cmd in "$here"/cases/*.cmd; do
+  name=$(basename "$cmd" .cmd)
+  # Each .cmd holds the commands to run after startup.
+  out=$( { printf 'set scroll continuous\n'; cat "$cmd"; printf 'exit\n'; } \
+         | (cd "$workdir" && timeout 60 "$binary" 2>&1) )
+
+  if printf '%s' "$out" \
+       | grep -qE 'runtime error|AddressSanitizer|SEGV'; then
```
benjub: You should also grep "?BUG CHECK". This would match the output of test case `assignvar-undeclared-var`, which should therefore remain "open".

Then, fix that bug in a separate issue/PR.

benjub (commented):
Two more comments:

Since databases with NUL are now (correctly) rejected, we should probably change mmpars.c:6220 from
```
    print2("?Error: file \"%s\" was not found\n", fullInputFn);
```
to
```
    print2("?Error: file \"%s\" could not be read\n", fullInputFn);
```
Else, you can read things like:
```
$ metamath < nul-in-source.cmd
?Sorry, the file "nul-in-source.mm" has a NUL character at byte 8.
?Error: file "nul-in-source.mm" was not found
Reading source file "nul-in-source.mm"... 0 bytes
?Source was not read due to error(s).  Please correct and try again.
```

---

A performance issue raised by Claude: you added
```
static flag lacksDummyVarMapping(long stmtNum)
```
which is called by proveFloating once per surviving candidate statement (mmpfas.c:1791), and it walks the candidate's entire reqHypList. On a clean database the answer is always 0 — but getting to 0 still means chasing reqHypList[hyp] into a scattered set of statement_structs. hasVarWithoutHyp is fixed once parsing completes, so the transitive property could be computed once per statement at the end of parseStatements and stored.

Then Claude proposed a fix: a global that short-circuits it, not a per-statement field:
In mmdata.h, beside the g_dummyVarBase declaration:
```
/*! Nonzero if any statement in the database has hasVarWithoutHyp set.  Lets
    the proof assistant skip the per-candidate hypothesis walk entirely on a
    database that parsed without errors.  See mmdata.c. */
extern flag g_anyVarWithoutHyp;
```
In mmdata.c, beside long g_dummyVarBase = 0;:
```
flag g_anyVarWithoutHyp = 0;
```
In mmpars.c at both places where one sets hasVarWithoutHyp (mmpars.c:1313 and mmpars.c:1710):
```
            g_Statement[stmt].hasVarWithoutHyp = 1;
            g_anyVarWithoutHyp = 1;
```
and clear it at the top of parseStatements next to g_dummyVarBase = g_mathTokens;, plus in eraseSource next to g_dummyVarBase = 0;.
In mmpfas.c, at the top of the helper:
```
static flag lacksDummyVarMapping(long stmtNum) {
  long hyp;
  if (!g_anyVarWithoutHyp) return 0; // No statement anywhere has the flag
  if (g_Statement[stmtNum].hasVarWithoutHyp) return 1;
  ...
```

I cannot evaluate the balance cost/benefit of that proposal. I leave it up to you to decide on whether to implement (something like) it, probably in a future PR, or not.

src/mmvstr.c
benjub: To be clear: I think siimply removing these 13 added lines and fixing L613 is better. The current fix can still trigger UBs when enough tabs are on a given line.

david-a-wheeler: I agree L613 is required.

HOWEVER: removing the allocation reintroduces two overflows: edit-untab-overflow and edit-untab-many-tabs both fail at mmvstr.c:596. The first is a single tab with "T" — the expansion loop runs under untab_flag || tab_flag, while i = i * 7 only fired for untab_flag. And i * 7 was never sufficient anyway: 40 tabs need 322 bytes, not 294.

There's also a third case neither version catches: CLEAN f "P,U" clears the parity bit, and '\211' & 0x7F is a tab — so tabs appear after any count is taken. An allocation of i * 8 doesn't depend on which characters are tabs. Added edit-parity-untab.

So: no to removal, yes to change.

Maybe many of the "tools" should just go away, but my current goal is just "get rid of UBs without impacting functionality".

src/mmvstr.c
david-a-wheeler: Agreed. Put in a commit.

src/mmpars.c
david-a-wheeler: I agree. Commit 51fd2b87be8bc8fbdd591ef59b8adb2400cb3631 will do that.

src/mmpars.c
david-a-wheeler: Commit 51fd2b87be8bc8fbdd591ef59b8adb2400cb3631 will do that.

src/mmpars.c
david-a-wheeler: Ah, I need to fix the comment. Will do.

src/mmpars.c
david-a-wheeler: It's more to provide airtight protection. I'll update the comment.

src/mmpfas.c
david-a-wheeler: Yes, they're the same "is this a dummy variable" test, and > g_dummyVarBase is the correct one. 

HOWEVER: It doesn't stop at mmpfas.c. mmunif.c documents > g_mathTokens as its contract for variables that can be assigned, and relies on it in about a dozen places — and mmpfas.c hands its schemes straight to unify(). If we change one side only and the two disagree about what a token means, that would be worse than the inconsistency we started with.

I'm trying to focus in this PR on "little repairs to counter UB". It's already getting MUCH longer than I intended.

I'd rather handle this as its own PR covering both files. It's a no-op on any database that parses cleanly.

david-a-wheeler: So my proposed resolution is "Work in a separate PR".

david-a-wheeler:
Just to repeat the note here:

> [src/mmpfas.c line 2887, 2026-07-31] Can you check if similar substitutions should be done on/around lines 839, 911, 1103, 2100, 2616, 2975, 3002, 3036 ? In ambiguous cases, a comment in the code may help.

I propose handling that in a separate PR. I recorded the issue (so it won't be lost) in issue https://github.com/metamath/metamath-exe/issues/191

src/mmvstr.c
david-a-wheeler: Agreed. I found another value dependent on it, making the argument even stronger. Done.

src/mmdata.h
david-a-wheeler: Done. Checked on 64-bit x86 and 64-bit ARM.

src/mmpars.c
david-a-wheeler: Agreed, done.

src/mmpfas.c
david-a-wheeler: I've pulled that into its own issue, to be resolved separately: https://github.com/metamath/metamath-exe/issues/191

src/mmvstr.c
david-a-wheeler: Done.

src/mmpfas.h
david-a-wheeler: Done. I later turned this into an inline function (`HIGHEST_MATH_TOKEN` looks like a constant, which it is nt).

david-a-wheeler:
I created issue #192 "bug(1741): sourceError() can't place the pointer it was handed" so we can handle that separately.

src/mmvstr.c
david-a-wheeler: See issue #192, bug(1741): sourceError() can't place the pointer it was handed

fuzz/run-cases.sh
david-a-wheeler: In separate issue for a sepratae PR.

src/mmvstr.c
david-a-wheeler: Done.

src/mminou.c
david-a-wheeler: Done

src/mminou.c
david-a-wheeler: Done

src/mmcmds.c
david-a-wheeler: I reworded it differently.

david-a-wheeler:
Whew! Resolved all (though in a few cases "resolved" was "create a new issue").

Anything else? I'd like to get this tome merged & eliminate this HUGE number of UBs, but of course, it's most important to "do no harm".

src/mmvstr.c
benjub: apparently not

benjub:
> Two more comments:
> 
> Since databases with NUL are now (correctly) rejected, we should probably change mmpars.c:6220 from
> ...

You fixed that in this PR.

> A performance issue raised by Claude: you added
> ...

What are your plans for that performance issue?

david-a-wheeler:
Regarding performance, there's nothing to win if I understand you correctly.

I took 137 theorems spread across set.mm (every 200th $p past statement 20000), each PROVEd, DELETE ALL, then IMPROVE ALL /DEPTH 1 — about 35 seconds of almost nothing but proveFloating. I compared this branch against a build with lacksDummyVarMapping() replaced outright by return 0;, which is the best case the proposal could ever reach, since it deletes the call's body rather than short-circuiting it, and did 2 runs:

This branch: 34.88 s, 35.00 s
Short-circuit:  37.72 s, 38.73 s

Removing the work doesn't increase speed (it's even slightly *slower*).

david-a-wheeler:
I think I know why it's no faster and in practice slower. I think that once again, memory caching behavior dominates performance.

The statement_structs are already in L1 by the time we get there. The call sits immediately after the loop at 1756-1786, which iterates the same reqHypList indices in the same order reading .type — and type and hasVarWithoutHyp are adjacent, both inside the first 64-byte cache line. It also only runs for candidates that already survived quickMatchFilter().

Against that, the global adds multiple places to keep in step, and any error makes things worse. It's not faster, and it's easy to get wrong... so I think it's better this way.

fuzz/run-cases.sh
```diff
@@ -0,0 +1,78 @@
...
+set -u
+
+here=$(cd "$(dirname "$0")" && pwd)
+binary=${1:-"$here/metamath-san"}
+
+if [ ! -x "$binary" ]; then
+  echo "$binary is not executable; run ./build-sanitizer.sh first" >&2
+  exit 2
+fi
+
```
benjub, line 21:
```suggestion

# The binary is invoked after cd'ing into $workdir, so a relative path
# would no longer resolve.  Make it absolute while we are still in $PWD.
case "$binary" in
  /*) ;;
  *) binary=$(cd "$(dirname "$binary")" && pwd)/$(basename "$binary") ;;
esac

```
Or something like that?

src/mminou.c
```diff
@@ -1430,19 +1516,13 @@ vstring readFileToString(const char *fileName, char verbose, long *charCount) {
...
-          (*charCount), strlen(fileBuf));
-    }
-  }
-/*E*/db = db + i; // For memory usage tracking (ignore stuff after null)
+  // Nothing here looks for a NUL inside the file any more.  There cannot be
+  // one: the scan further up returns NULL for any file that has one.  So
+  // strlen(fileBuf) is *charCount, unless bug(1522) just above has fired, in
+  // which case the count is already wrong in ways this line cannot mend.
+  // What stood here warned about a NUL and then handed the truncated buffer
+  // back regardless, which is the behavior that scan replaced.
```
benjub: Why comment older versions of code? I'd remove that.

benjub:
Two hopefully final comments: one on run-cases.sh, and one on your new commit.

benjub:
> Two hopefully final comments: one on run-cases.sh, and one on your new commit.

@david-a-wheeler hopefully you can act on these two comments and merge. I have an MR on top of this one with further UB fixes.

david-a-wheeler:
Sorry for the delay, I had some other rush work! I'm going to try to get to this soon...!

david-a-wheeler:
BTW, any thoughts on *how* I merge once these last issues are addressed? If you're building on top, then the obvious way is a merge pull requests, which shows all of the individual commits. That's a lot of commits at once, but then you can see the history of each commit.

The alternative is a squashed merge. I think there are so many changes here that the squashed merge would be too hard to read. But let me know if you disagree.

fuzz/run-cases.sh
david-a-wheeler: Good catch, thank you.  Taken, though I built the path a different way.

It is worse than the binary not resolving: the failure is silent.  The
`-x` check runs in the caller's directory, so it passes.  The exec then
fails inside `$workdir`, and the error goes to `$out` through the
`2>&1`, where it matches none of the grep alternatives.  So the case
reads as clean.  With a stub binary that does nothing but print
`?BUG CHECK`:

```
$ sh fuzz/run-cases.sh fuzz/metamath-san    # relative path
ok          demo
$ echo $?
0
```

For a harness whose whole premise is that clean means fixed, that says
the opposite of what happened, which is the same trouble as the
`?BUG CHECK` case you raised earlier.

I used `$PWD` rather than `cd`, `dirname` and `pwd`:

```sh
case $binary in
  /*) ;;
  *) binary=$PWD/$binary ;;
esac
```

The reason is that `cd` consults `CDPATH`, and when it uses it, it
prints the directory it picked on stdout, straight into the command
substitution.  With `CDPATH` set in the environment the suggested form
puts a two line string in `$binary`, and the quiet pass comes back:

```
CDPATH set, binary given as a relative path:
  cd/dirname/pwd :  ok          demo             exit 0    <- quiet pass
  $PWD           :  REGRESSION  demo: DETECTED   exit 1
```

That is not one shell being odd; dash, bash and busybox sh all echo it.
We want an absolute path here rather than a canonical one, and `.` and
`..` resolve fine at exec time, so the prefix does the whole job and
there is no subshell.  I kept your placement after the `-x` check, so
the error message still names the path the user typed.

Two more things came out of this, both the same quiet failure.

A case that hangs was also reported ok, since `timeout` kills it and the
partial output matches nothing either.  The script now looks at the exit
status of the run.  The test is "not 0 or 1" rather than "is 124",
because that number is not portable: GNU and BSD `timeout` exit 124,
busybox exits 143.  metamath itself exits 0 or 1 and nothing else, so
anything above 1 means the run did not finish.  The grep still runs
first, so a run that produced a sanitizer report is labeled `DETECTED`
as before.

I didn't *test* on MacOS/Homebrew, but I did a check.
macOS ships no `timeout` at all, so this suite has never run there:
every case printed ok and the script exited 0.  Homebrew's coreutils
installs it as `gtimeout`, so the script looks for both names now and
exits 2 with a clear message when neither is there.

The 28 case suite gives byte identical output before and
after, and a relative invocation now agrees with an absolute one.  A
nonexistent binary still exits 2.  Paths holding spaces work, a hung
case reports "did not finish (status 124)" and exits 1, and as an
`open-` case it stays known-open rather than being announced as fixed.
`shellcheck -s sh` is clean.

src/mminou.c
david-a-wheeler: Fair enough! I agree. I'll add a commit to fix that.

david-a-wheeler:
@benjub - all okay? Ready to merge this (as a merge pull request, *not* as a squash merge)?

I normally don't create a PR with this many changes. When I added the fuzzer it revealed a lot, and it made since to fix them in a PR... at least, it did when I started.

NOTE TO OTHERS: These UBs would *NOT* have made it easy to slip something into set.mm. The set.mm database undergoes proof-checking by *many* different tools in *many* different languages, so a UB in metamath-exe is *not* enough to get accepted. @benjub knows this, but perhaps other readers here don't.

benjub (approved):

benjub:
> @benjub - all okay? Ready to merge this (as a merge pull request, _not_ as a squash merge)?

Yes, merge pull request. Thanks!

david-a-wheeler:
Merged!

benjub:
@david-a-wheeler can we delete merged branches for this repo?

david-a-wheeler:
Done. Deleting branches doesn't really do much, though. If you clone a git repo, unless you limit depth you get *all* changes, so you'd still get those commits in the repo via the main (master) branch.

PR #194, opened 2026-08-05
describe invocation of regression tests

wlammen:
Documentation is an ongoing effort, and each state is (likely) better than the one before, but still intermediate. Don't hesitate to review and merge the current version, subsequent pull requests can still apply modifications on the side.  If I think a version is too bad for acceptance, I will convert the pull request temporarily into a draft.

These are the objectives for this pull request:

1. rename macro RUN_TESTS to RUN_TESTS_AND_EXIT_IF_ENABLED for better understanding
2. give a detailed description in a doxygen formatted comment of how to run regression tests

Both the executables **metamath** and **metamath_test** can be created on my computer and run as expected.
The reworked documentation looks like this on a doxygen generated HTML page:

grafik

### DIFF
--- a/src/metamath.c
+++ b/src/metamath.c
@@ -735,7 +735,6 @@ int main(int argc, char *argv[]) {
 
   // see mmtest.h for how to run the regression tests
   RUN_TESTS_AND_EXIT_IF_ENABLED();
-  // The following code is not reached if tests are enabled.
 
   // ******** If g_listMode is set to 1 here, the startup will be Text
   //          Tools utilities, and Metamath will be disabled *********
--- a/src/metamath.c
+++ b/src/metamath.c
@@ -731,12 +731,11 @@ void command(int argc, char *argv[]);
  */
 int main(int argc, char *argv[]) {
 
-// argc is the number of arguments; argv points to an array containing them
+  // argc is the number of arguments; argv points to an array containing them
 
-#ifdef TEST_ENABLE // enable this in mmtest.h or via './build.sh -t'
-  RUN_TESTS();
-  // you never get here
-#endif
+  // enable this in mmtest.h or via './build.sh -t'
+  RUN_TESTS_IF_ENABLED();
+  // The following code is not reached if tests are enabled.
 
   // ******** If g_listMode is set to 1 here, the startup will be Text
   //          Tools utilities, and Metamath will be disabled *********
--- a/src/metamath.c
+++ b/src/metamath.c
@@ -734,7 +734,7 @@ int main(int argc, char *argv[]) {
   // argc is the number of arguments; argv points to an array containing them
 
   // see mmtest.h for how to run the regression tests
-  RUN_TESTS_IF_ENABLED();
+  RUN_TESTS_AND_EXIT_IF_ENABLED();
   // The following code is not reached if tests are enabled.
 
   // ******** If g_listMode is set to 1 here, the startup will be Text
--- a/src/metamath.c
+++ b/src/metamath.c
@@ -733,7 +733,7 @@ int main(int argc, char *argv[]) {
 
   // argc is the number of arguments; argv points to an array containing them
 
-  // enable this in mmtest.h or via './build.sh -t'
+  // see mmtest.h for how to run the regression tests
   RUN_TESTS_IF_ENABLED();
   // The following code is not reached if tests are enabled.
 
--- a/src/mmtest.h
+++ b/src/mmtest.h
@@ -29,15 +29,15 @@
    ./metamath_test
    \endverbatim
  *
- * 4. The macro \ref TEST_SILENT can suppress progress and success information.
- * When left in its default state, messages of the following kind are issued:
+ * 4. The test program reports its progress to \c stdout:
  * \verbatim
    running test_mmfatl:test_fatalErrorInit... ok
    \endverbatim
  *
  * If every displayed test ends with \c ok, all regression tests have passed.
- * Otherwise, failing tests always display diagnostic information identifying
- * the test and the kind of failure.
+ * Progress and success messages can be suppressed, see \ref TEST_SILENT.
+ * Failing tests always display diagnostic information identifying the failing
+ * assertion and its source location.
  *
  * \subsection test_details Details and implementation
  * If the macro \c TEST_ENABLE is defined by the \c -t option of \c build.sh,
@@ -55,11 +55,9 @@
  * executable does not increase in size. Thus, a disabled test suite incurs
  * neither a linking nor a runtime penalty.
  *
- * If testing is enabled, the progress of tests is reported to \c stdout,
- * subject to \ref TEST_SILENT. The first failed assertion aborts the test
- * suite to which it belongs, and a diagnostic message identifying the failing
- * assertion and its source location is always written. Remaining test suites
- * still run, and the program exits with a nonzero status if any test failed.
+ * A failed assertion skips the rest of the function in which it occurs. The
+ * other functions registered with \c RUN_TEST are still executed. The program
+ * exits with status 0 if all tests passed, and nonzero otherwise.
  *
  * We recommend running the regression tests whenever the code is modified
  * to ensure that it continues to behave as intended. The tests are also
--- a/src/mmtest.h
+++ b/src/mmtest.h
@@ -43,7 +43,7 @@
  * the macro is defined matters. Defining the macro directly in this file does
  * not affect the executable's name.
  *
- * If testing is enableded or disablrd without modifying any source files in
+ * If testing is enableded or disabled without modifying any source files in
  * between, also pass the \c -c (clean) option to \c build.sh to remove
  * intermediate build artifacts.
  *
--- a/src/mmtest.h
+++ b/src/mmtest.h
@@ -38,12 +38,12 @@
  * If every test ends with \c ok, all regression tests have passed.
  * Otherwise, diagnostic information identifies the failed test.
  *
- * If the macro \c TEST_ENABLE is defined (the \c -t option of
- * \c build.sh), the regression tests are compiled into a separate
- * executable named \c metamath_test. The value of \c TEST_ENABLE can
- * also be overridden in this file. If testing is switched on or off
- * without modifying any source files in between, also pass the \c -c
- * (clean) option to \c build.sh to remove intermediate build artifacts.
+ * If the macro \c TEST_ENABLE is defined (the \c -t option of \c build.sh),
+ * the regression tests are compiled into a separate executable named
+ * \c metamath_test. The macro may also be defined directly in this file; its
+ * value is irrelevant. If testing is switched on or off without modifying any
+ * source files in between, also pass the \c -c (clean) option to \c build.sh
+ * to remove intermediate build artifacts.
  *
  * If testing is disabled, the \c RUN_TESTS_IF_ENABLED macro expands to
  * nothing. The compiler also excludes all test code, so the resulting
--- a/src/mmtest.h
+++ b/src/mmtest.h
@@ -43,9 +43,9 @@
  * the macro is defined matters. Defining the macro directly in this file does
  * not affect the executable's name.
  *
- * If testing is enableded or disabled without modifying any source files in
- * between, also pass the \c -c (clean) option to \c build.sh to remove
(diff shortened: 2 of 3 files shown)


src/mmtest.h
```diff
@@ -7,42 +7,77 @@
...
- * Controls whether the regression tests for a
- * particular module is in/excluded.
+ * If testing is disabled, the \c RUN_TESTS_IF_ENABLED macro expands to
+ * nothing. The compiler also excludes all test code, so the resulting
+ * executable does not increase in size. Thus, a disabled test suite incurs
+ * neither a linking nor a runtime penalty.
+ *
+ * If testing is enabled, the tests report their progress to \c stdout.
+ * Execution stops at the first regression failure, and a diagnostic
+ * message provides further details about the context of the failure.
```
benjub:
```suggestion
 * The first failed assertion aborts the test in which it occurs, and a
 * diagnostic message identifies the failing assertion and its source
 * location.  Remaining tests still run, and the program exits with a
 * nonzero status if any test failed.
```

src/mmtest.h
```diff
@@ -7,42 +7,77 @@
...
- * macro, no value, just defined or not.
+ * 3. Execute the following commands:
+ * \verbatim
+   ./build.sh -ct
+   ./metamath_test
+   \endverbatim
+ *
+ * 4. The test program produces output similar to the following:
+ * \verbatim
+   > running test_mmfatl:test_fatalErrorInit... ok
```
benjub:
```suggestion
   running test_mmfatl:test_fatalErrorInit... ok
```

src/mmtest.h
```diff
@@ -7,42 +7,77 @@
...
+ *  RUN_TESTS_IF_ENABLED();
+ *
+ *  // This code is not reached if tests are enabled.
+ *
+ *  // Code performing normal execution of Metamath goes here.
+ * }
+ * \endcode
  */
 
+
```
benjub:
```suggestion
```
The rest of the files do not use double blank lines.

src/mmtest.h
```diff
@@ -7,42 +7,77 @@
 #ifndef METAMATH_MMTEST_H_
 #define METAMATH_MMTEST_H_
 
+
```
benjub:
```suggestion
```
The rest of the files do not use double-blank lines.

src/metamath.c
```diff
@@ -731,12 +731,11 @@ void command(int argc, char *argv[]);
...
 int main(int argc, char *argv[]) {
 
-// argc is the number of arguments; argv points to an array containing them
+  // argc is the number of arguments; argv points to an array containing them
 
-#ifdef TEST_ENABLE // enable this in mmtest.h or via './build.sh -t'
-  RUN_TESTS();
-  // you never get here
-#endif
+  // enable this in mmtest.h or via './build.sh -t'
```
benjub:
```suggestion
  // see mmtest.h for how to run the regression tests
```
To avoid drifting from what mmtest.h says. (It is already drifting since you wrote there "build.sh -ct".)

src/mmtest.h
```diff
@@ -7,42 +7,77 @@
...
+   \endverbatim
+ *
+ * 4. The test program produces output similar to the following:
+ * \verbatim
+   > running test_mmfatl:test_fatalErrorInit... ok
+   \endverbatim
+ *
+ * If every test ends with \c ok, all regression tests have passed.
+ * Otherwise, diagnostic information identifies the failed test.
+ *
```
benjub, line 41:
```suggestion
 * This assumes the default value of \ref TEST_SILENT; if it is \c true,
 * then successful tests stay silent and only failures produce output.
 * In either case, \c metamath_test exits with status 0 if all tests passed,
 * and nonzero otherwise.
 *
```

benjub (commented):
I made a few comments in the text.

You should probably add to `CONTRIBUTING.md`, after "We expect all code merged to compile and not break existing functionality.", something like:

```
We also expect the regression tests to pass.  See `src/mmtest.h` for how to
run them and for more details.
```

src/mmtest.h
wlammen: I am not convinced to document the behavior of the program twice, as some information is repated in a later paragraph.

benjub: You're right: partly redundant. I'm going to propose a consolidated paragraph.

src/mmtest.h
```diff
@@ -9,38 +9,77 @@
...
- * \def TEST_ENABLE
- * macro, no value, just defined or not.
+ * 3. Execute the following commands:
+ * \verbatim
+   ./build.sh -ct
+   ./metamath_test
+   \endverbatim
+ *
+ * 4. The macro \ref TEST_SILENT can suppress progress and success information.
+ * When left in its default state, messages of the following kind are issued:
```
benjub:
```suggestion
 * 4. The test program reports its progress to \c stdout, one line per test:
```

src/mmtest.h
```diff
@@ -9,38 +9,77 @@
...
+ * \verbatim
+   running test_mmfatl:test_fatalErrorInit... ok
+   \endverbatim
+ *
+ * If every displayed test ends with \c ok, all regression tests have passed.
+ * Otherwise, failing tests always display diagnostic information identifying
+ * the test and the kind of failure.
```
benjub:
```suggestion
 * Setting \ref TEST_SILENT to \c true suppresses these progress and success
 * messages. Failing tests always display diagnostic information identifying
 * the failing assertion and its source location.
```
This and the previous edits are merely putting the TEST_SILENT remark after having shown the displayed output. Also, the diagnostic information is about "the failing assertion", which is more precise than "the test".

src/mmtest.h
```diff
@@ -9,38 +9,77 @@
...
+ * If testing is disabled, the \c RUN_TESTS_AND_EXIT_IF_ENABLED macro expands
+ * to nothing. The compiler also excludes all test code, so the resulting
+ * executable does not increase in size. Thus, a disabled test suite incurs
+ * neither a linking nor a runtime penalty.
+ *
+ * If testing is enabled, the progress of tests is reported to \c stdout,
+ * subject to \ref TEST_SILENT. The first failed assertion aborts the test
+ * suite to which it belongs, and a diagnostic message identifying the failing
+ * assertion and its source location is always written. Remaining test suites
+ * still run, and the program exits with a nonzero status if any test failed.
```
benjub: First sentence is redundant with the paragraph above. Second sentence seems false: only the current test aborts, not the test suite. Also, be more precise about exit status. Suggestion:
```suggestion
 * A failed assertion aborts only the test in which it occurs; the remaining tests
 * still run. The program exits with status 0 if all tests passed, and nonzero
 * otherwise.
```

benjub (commented):
Hopefully the last two comments. The rest is good for me.

When merging, we'll probably squash, and edit the two commit messages:
- invokation --> invocation
- reove --> remove

src/mmtest.h
wlammen: It's actually not a line per test, but a line issued after executing a suite of tests bundled in a runTest call. Future implementations of regression tests may also protocol more and other intermediate test results (hints and so on). One should not be too concrete here, and I prefer the original version, perhaps stdout could be added, if it is not clear from the beginning.

Perhaps one should write:
4. Progress information on \c stdout can optionally be suppressed. By default, messages of the following kind are reported:

src/mmtest.h
wlammen: How TEST_SILENT is properly set and for what effect, should be part of the documentation of TEST_SILENT, and not the file description outlining the general concepts of regression tests and their invocation from the perspective of a developer curious about whether he has introduced bugs. Replacing test with assertion is ok.

src/mmtest.h
wlammen: The ASSERTF macro expands to code that finishes a failed test with a `return false;`. This exits the currently running function, and later tests encoded in it are not executed. This is on purpose:

   On failure a test writes a
   diagnostic message to \c stdout and returns false.  This skips all
   following tests, assuming that basic tests are executed first. If these
   already fail, tests depending on those basics are likely to fail, as
   well. Avoid their distracting failure messages.

As for the duplication, I don't like it, and would have dropped it myself. The problem is, that the same information appears in two different sections. One is a sort of help text, and allows a user to interpret the output on the screen. The second section is about concepts, and highlights the dependencies on optionally activated features.

wlammen:
Behind the scenes I meanwhile reworked more comments, and added missing ones.  How should I proceed without overloading the workflow?

src/mmtest.h
benjub: The sentence about TEST_SILENT is basically the one that was a few lines above at the top of section 4.
I think that information belongs to "how to run the tests".

src/mmtest.h
benjub: Then, 
```suggestion
 * 4. The test program reports its progress to \c stdout:
```
but do not begin a part with documenting an optional behaviour: that comes after the nominal output is given to the reader.

src/mmtest.h
benjub:
```suggestion
 * Progress and success messages can be suppressed, see \ref TEST_SILENT.
 * Failing tests always display diagnostic information identifying the failing
 * assertion and its source location.
```

src/mmtest.h
benjub: I see. We didn't have the same terminology, but I agree on the message. What about
```suggestion
 * A failed assertion skips the rest of the function in which it occurs. The other
 * functions registered with \c RUN_TEST are still executed. The program exits
 * with status 0 if all tests passed, and nonzero otherwise.
```
(I want to keep the clear exit code description.) You may add your explanation ("If these already fail...") to the doc of ASSERTF. By the way: the docstrings for ASSERT and ASSERTF lack the initial special character /*! it seems.

I hope you agree on that suggestion: a failed assertion does NOT stop everything: an example gave me:
```
$ ./build.sh -ct
running test_mmfatl:test_unsignedToString...
test_unsignedToString: assertion 1 == 2 failed at src/mmfatl.c:668
TEST test_mmfatl:test_unsignedToString FAILED
running test_mmfatl:test_fatalErrorInit... ok
running test_mmfatl:test_isBufferOverflow... ok
... all 11 remaining tests ... ok
exit=1
```

benjub:
> Behind the scenes I meanwhile reworked more comments, and added missing ones. How should I proceed without overloading the workflow?

Not in this PR please, which I hope we can merge soon. Focused PRs are easier to review.

For the bigger picture, @david-a-wheeler suggested in #188 that the discussions on new projects move to a new issue (or issue**s**), which seems right to me.  I do not have much bandwidth, and given the slow rate of PR merges, I think that bugfixes and incremental cleanups will have more chance to get reviewed than larger projects (unicode, reference counting), which risk not finding reviewers with enough time.

src/mmtest.h
wlammen: This PR is only about the file description and the name of one macro, see my introductory comment.  I deliberately limited the changes to these two elements, to learn how to contribute in an efficient way, and to avoid a long awaiting of an approval.

Perhaps one should explain somewhere what a test suite is: a series of tests, the most primitives and independent ones first, followed by those depending on the ones before. In such a context, aborting the suite makes mostly sense.

In general, I would like to have a high throughput.  This means minor quality issues can be ignored. A later pull request may correct them, or even you may create a separate PR addressing individual points hurting you much.

wlammen:
If nothing else helps, I will create my personal fork.

benjub (approved):
Good for me. @david-a-wheeler can you have a look and squash-and-merge?

As for @wlammen's other questions regarding process, I think @david-a-wheeler is also in a better position to reply.

@wlammen : On defining "test suite": the file uses the term once, in "a disabled test suite incurs neither a linking nor a runtime penalty", where it means the whole test machinery, while your definition describes the assertion sequence inside one RUN_TEST function. I do not know what is best; this can indeed be done in a follow-up PR.

wlammen:
I updated the screenshot of the documentation in the introductory comment to reflect the latest changes.

benjub:
@david-a-wheeler can you have a look at the PR and Wolf's questions regarding process?  It'd be nice to merge this PR soon.

david-a-wheeler:
Sorry, I *just* opened this PR to look at it. But I expect that it's fine, and if not, we'll open another PR :-).

PR #195, opened 2026-08-06
Fix some UBs, harden mmfatl.c and use it

benjub:
- On top of #190.
- Best reviewed by commits; see commit messages.
- Design decisions mine; assisted by AI for some code and commit messages.

I requested @david-a-wheeler as reviewer since it's in the continuation of UB fixing and #190. 

@wlammen can you take a look at the mmfatl.c/h changes ?


src/mmfatl.c
```diff
@@ -855,21 +895,30 @@ static bool test_fatalErrorPrintAndExit(void) {
   return true;
 }
 
+// Exercises composeErrorAt(): everything fatalErrorExitAt() does apart from
+// emitting the message and terminating.
+static void testcase_ExitAt(char const* file, unsigned line,
```
wlammen, line 925: There is still a problem with the file argument, if it was generated by the `__FILE__` macro.  The file names in Metamath are pure ASCII, but that need not hold for the path.  On usual systems these may contain UTF-8 elements.  On exotic systems these can be neither ASCII nor UTF-8. Okay, one can ignore the exotic case relatively safely nowadays, but a strict solution might check for the  execution character set first, before using `__FILE__` macro.  Or since you rely on GNU anyway, use its better suited FILE_NAME instead. If you want to allow UTF-8 in general, the composeErrorAt function should care about truncated UTF-8 sequences due to buffer overflow.

src/mmfatl.c
```diff
@@ -219,12 +219,13 @@ struct ParserState {
...
-   * guaranteed on invocation by the caller.
+   * guaranteed on invocation by the caller.  The list itself is not held here:
+   * it belongs to the call, not to the parser, so it is threaded through
+   * \ref parse and \ref handleSubstitution as a parameter.
    */
-  va_list args;
 };
 
-static struct ParserState state;
+static struct ParserState state = { .out = &buffer, .format = "" };
```
wlammen, line 228: Is this C99?

src/mmfatl.c
```diff
@@ -425,54 +426,72 @@ bool fatalErrorPush(char const* format, ...) {
...
 
   return !overflow;
 }
 
-void fatalErrorPrintAndExit(void) {
+/*!
+ * \brief finish the message in the buffer, without emitting it.
+ *
+ * The part of \ref fatalErrorPrintAndExit that carries logic: it pads a
+ * non-empty message with a LF when it lacks one.  Split out so the regression
```
wlammen: One could add here why a LF is padded to the right. The end of a message might appear with start of the next one, or a command prompt after exit() on the same line.

src/mmfatl.c
benjub: Yes, it is C99 (but not C90 indeed): `gcc -std=c99 -pedantic-errors -Wall -Wextra mmfatl.c` compiles. Initialization of a static by constant expressions or string literals (&buffer qualifies since buffer is declared above as static).

Your remark made me notice that since `format` is const, one can streamline `initState` to have
`state->format = "";`. Done in https://github.com/metamath/metamath-exe/pull/195/commits/0e69d51b7a929e44c8e2caf8ca420759ce3a117b.

src/mmfatl.c
benjub: Done in https://github.com/metamath/metamath-exe/pull/195/commits/897273efaba6b80a110e4c3c8912f9cb26a4bc13.

src/mmfatl.c
benjub: You're right that the path can be non-ASCII. But it isn't a correctness problem for mmfatl (it never interprets bytes as characters). C99 gives the preprocessor no way to query the execution character set.

I used your `__FILE_NAME__` suggestion (see https://github.com/metamath/metamath-exe/pull/195/commits/d415b8a1fb9f050f2db166ecac517faa4ace67e1), via a macro.  I did it not for the encoding but because dropping the directory removes both the only part that can be non-ASCII and the build-path leak, and gives the message more buffer space (since the path may be long).

On truncated UTF-8: I think this is not an issue, and the handling of one glyph is not worth the trouble and the maintenance risk to add complexity to mmfatl.c, which is meant to work with a corrupt heap and no locale. With `__FILE_NAME__`, it becomes even less necessary.

wlammen: I think, supporting UTF-8 in error messages is an invest into the future, even if the effect is not great right now.  May I suggest to add the following code to mmfatl.c:

/*!
 * \brief Check whether the message buffer contains any text.
 *
 * \param[in] buffer [not null] Buffer to check.
 *
 * \return
 *   \c true if no text has been appended to the buffer; otherwise
 *   \c false.
 *
 * \pre
 *   \ref initBuffer has been called for \p buffer.
 */
static inline bool isBufferEmpty(struct Buffer const *buffer)
{
  return buffer->begin == buffer->text;
}

/*!
 * \brief Return the last character stored in the message buffer.
 *
 * Returns the character immediately preceding the current insertion position
 * in \p buffer. Characters discarded because of a buffer overflow are not
 * considered part of the message. If the message buffer is empty, \c NUL is
 * returned.
 *
 * \param[in] buffer [not null] The message buffer to inspect.
 *
 * \return The last character stored in \p buffer, or \c NUL if the message
 * buffer is empty.
 *
 * \pre \ref initBuffer has been called for \p buffer.
 */
static char getLastBufferedChar(struct Buffer const *buffer)
{
  return isBufferEmpty(buffer) ? NUL : *(buffer->begin - 1);
}

/*!
 * \brief Delete the last character stored in the message buffer.
 *
 * If the message buffer is not empty, removes the character immediately
 * preceding the current insertion position and replaces it with \c NUL.
 * If the buffer is full, it is no longer full after the deletion.
 *
 * If the message buffer is empty, the function has no effect.
 *
 * \param[in,out] buffer [not null] The message buffer to modify.
 *
 * \pre \ref initBuffer has been called for \p buffer.
 *
 * \post If \p buffer was not empty, its insertion position is moved back
 * by one byte and the deleted character is replaced with \c NUL.
 */
static void deleteLastBufferedChar(struct Buffer *buffer)
{
  if (!isBufferEmpty(buffer))
    *--(buffer->begin) = NUL;
}

/*!
 * \brief Remove an incomplete UTF-8 sequence from the message buffer.
 *
 * A UTF-8 character may occupy up to four bytes. If the message is
 * truncated, a multibyte character may therefore be cut off before its
 * complete encoding has been written. This function removes trailing bytes
 * that cannot occur in an ASCII character, thereby discarding the
 * incomplete remainder of a possibly truncated UTF-8 character.
 *
 * In UTF-8, ASCII characters are encoded as single bytes in the range
 * 0x00 through 0x7F, whereas continuation and multibyte leading bytes have
 * values above 0x7F. Consequently, the function has no effect on a buffer
 * containing only ASCII text.
 *
 * \param[in,out] state [not null] The parser state to update.
 * \return The number of bytes removed from the end of the message buffer.
 *
 * \pre \ref initState has been called for \p state.
 * \post Any incomplete UTF-8 sequence at the end of the message buffer has
 * been removed.
 */
static unsigned deleteCutoffUtf8(struct ParserState *state)
{
  unsigned deleted = 0;

  while (getLastBufferedChar(state->out) < 0) {
    deleteLastBufferedChar(state->out);
    ++deleted;
  }

  return deleted;
}

benjub: I'd rather not add that complexity to `mmfatl.c`: the risk of new bugs isn't worth it, and I think your proposal shows why. I asked claude to try it and it found the following:
"""
It removes too much: the loop deletes complete multi-byte characters as well, so "café" becomes "caf" and an all-Cyrillic message becomes empty. The < 0 test also does nothing where plain char is unsigned (ARM, -funsigned-char), including in the case it's meant to fix. And deleting a byte clears the overflow state, so fputs stops before the ... — visible truncation becomes silent.
"""

But even if we fixed those, I think that encoding logic is too much for a module that has to stay minimal and work even in corrupt states. Metamath files are ASCII by spec and with `__FILE_NAME__`, we do not dsplay the path anymore, so we only have ASCII character. So you are trying to solve a hypothetical problem (which is not even that bad), while the extra cost is real.

wlammen: The first lead byte of an UTF-8 sequence has its total length encoded, by how many 1 bits are set top down.  I got this wrong in my quick and dirty suggestion. Routines like getLastBufferedChar and isBufferEmpty are useful, though, since they translate a particular parser state/operation into something easily understandable and verifiable. In an OOP implementation of ParserState they would appear as natural static methods.  But I am not insisting on this.

Generally spoken, the implementation of mmfatl.c was flawed: It could receive UTF-8 sequences from the FILE macro, but served this situation only to a 99% degree.  It could become cumbersome again at any time, if someone unaware of this flaw adds a source file with a non-ASCII name.

david-a-wheeler:
Thanks for doing this!

Now that I've merged #190, I'll ask an AI to review each of these commits (since they build on 190). I'll then examine any issues with the issues it finds. My concern when fixing UBs is always that we'll accidentally break functionality while fixing UBs, so I think it's important to do a double-check of this stuff.

david-a-wheeler:
Here's the review of this PR from `Claude:claude-opus-5`. No *big* problems! However @benjub you might look at the "smaller points" below to see if you want to address them.

NOTE: the build.sh "doesn't rebuild" problem is pre-existing & I believe we should *NOT* fix that in this PR. My cosmopolitan + wasm build PR changes a lot about building, let's merge *that* in first, and then address the problem if it still remains after that.

====

Source: <https://github.com/metamath/metamath-exe.git>, `refs/pull/195/head`,
fetched 2026-08-10.  Author: @benjub.  Reviewed against `master` at 4051b30,
which already carries PR #190.

The 15 commits below are the ones not in master.  PR #195 branched from
1ae47f6, so `git log master..pr195` is exactly its own work; nothing from
#190 is counted twice.

What I was watching for: these are UB fixes, and a UB fix that quietly
changes what the program computes is worse than the UB.  So the question at
every commit was not "is this UB" but "can a user see a different answer".

## Bottom line

All 15 are worth taking.  I found no case where the program computes a
different result.  There are six places where a user can see different
*text* or a different *exit status*, all of them on fatal or error paths,
and all of them improvements.  They are listed together below so they can go
in the release notes rather than surprise somebody.

Two things to raise with benjub, neither a blocker: `bug(1305)` now reports
with `printf()` instead of `print2()`, which takes it out of the session log;
and two commits landed with no explanation in the body.

## Evidence

Everything here was run on the merge of pr195 into master, not on pr195
alone, since that is what would land.

- The merge is clean.  No conflicts, and the merged tree builds with zero
  warnings under the project's own `./build.sh`.
- 39 of 39 tests pass.  The fuzz case suite is 28 ok and 1 known-open.
- The regression build passes: 12 of 12 mmfatl unit tests.  See the note on
  `build.sh -t` at the end; my first run of these was a false pass.
- **A full set.mm run is byte identical.**  READ of the 51 MB set.mm, VERIFY
  PROOF *, VERIFY MARKUP *, SHOW STATEMENT ax-mp /FULL, SHOW USAGE ax-mp and
  SHOW TRACE_BACK idi, 707 lines of output, master against the merge: the
  only differing line is "All proofs in the database were verified in NN s",
  and master differs from itself the same way on a second run.  Masking the
  elapsed time, the two are identical byte for byte.

That last one is the answer to "did we break anything".  On the workload the
program exists to run, it produces the same output it did before.

## Commit by commit

Oldest first.

**182539b, Fix UB in bug() and let the compiler check print2 formats.**
Accept.  `bugNum` is `int` and was printed through `%ld`.  Checked the two
other integers this touches: `bufsiz` is `int` and `charsPrinted` is `long`,
so `%d` and `%ld` land on the right ones now.  `%ld` and `%d` render a bug
number identically, so no output moves.  The format attribute is the real
value here: it is what found everything in the next two commits, and it is
what will catch the next one.

**4ee5c9b, print the file name, not the FILE pointer.**  Accept.  Two
genuine bugs: `fp` passed to `%s` when it is a `FILE *` and, on that branch,
NULL.  Output changes for the better, from `?Error trying to write "(null)"`
to the actual name.  No test captured the old text.

**d8acef8, pass g_fullArg elements to print2 as vstring.**  Accept.  The
casts are the correct half of gcc's advice: these elements are vstrings, and
following the `%p` hint would have printed addresses where file names belong.
That would have been a "fix" that broke output, so it is worth saying that
benjub went the other way.  I sampled the sites rather than reading all 21;
the warning-free build is the completeness argument, since the attribute is
what flags them.

Three changes in it are not UB and deserve their own look, since all three
touch what gets printed:

- `print2("...SIZE=-2>\n", "%")` loses the stray `"%"`.  Excess arguments to
  a format are evaluated and ignored, so nothing changes.
- `print2("")` becomes `print2("%s", "")`, to dodge -Wformat-zero-length.
  Both produce nothing.
- `print2(cat(...))` becomes `print2("%s", cat(...))`.  This one is a real
  latent bug being closed: the text was being used as a *format*, so a `%` in
  it would have been read as a conversion.  It cannot happen today, because
  the text is built from a statement label and a byte count and labels cannot
  hold `%`, so output is unchanged.  Worth taking anyway.

**deef140, check mmfatl format strings at compile time.**  Accept.  It
carries two real defect fixes: `state.out` was NULL until `fatalErrorInit()`
ran, so a `fatalErrorPush()` before that dereferenced NULL, and
`handleText()` wrote to the file-scope buffer instead of `state->out`.  The
second is a no-op in the program as it stands, since `state->out` is always
`&buffer` there, but it is the right shape.

The part worth checking is the rewrite of the error location.  It drops the
`"%sIn line %u:\n"` form, whose leading `%s` swallowed a file argument that
is NULL or empty on that branch.  That is safe: `handleSubstitution()` skips
a NULL argument entirely, so the old form already produced exactly
`In line N:`.  The unit test asserts that text.

One nit: it also reformats `#if defined __WATCOMC__` to
`#if defined (__WATCOMC__)` and drops blank lines in `fSafeOpen()`, which has
nothing to do with format strings.  Harmless, but it is noise in a commit
that is otherwise easy to check.

**fefc894, terminate in every build, stop storing the argument list.**
Accept, and the `va_list` half is more than tidying.  A `va_list` living in
a file-scope struct that outlives the call is a hazard in its own right, and
threading it through `parse()` and `handleSubstitution()` as a parameter also
removes the `va_copy`/`va_end` pair.  Splitting `finishMessage()` and
`composeErrorAt()` out so the tests can exercise the logic without the
`exit()` is the right way round: the shipped path no longer has a
test-only `#ifndef TEST_ENABLE` in it, so the thing tested is the thing that
runs.

**0f0c069, report fatal conditions in bug()'s dependency layer without
bug().**  Accept.  The reasoning is sound: `bug()` reports through
`print2()`, builds messages with `let()`, `cat()` and `str()`, and reads its
answer through `cmdInput1()`, so a `bug()` from any of those recurses to a
stack overflow and the user never sees the diagnostic.  `fatalErrorExitAt()`
allocates nothing, which is what makes it usable from the out-of-memory path;
I checked that mmfatl contains no `malloc`, `let` or `cat`.

Confirmed the two bug-check spellings produce identical text, which matters
because `fuzz/run-cases.sh` greps for it: `BUG_CHECK_FORMAT` with `%d` and
`BUG_CHECK_FATAL_FORMAT` with `%u` both give
`?BUG CHECK:  *** DETECTED BUG 1741`.

Output does change at these sites: the bug-check line now comes first and the
`*** FATAL ERROR ***` line second, where `pushTempAlloc()` printed them the
other way round, and an `At file:line` line is added above both.

**3b1a79f, remove the bug() recursion through print2()'s back buffer.**
Accept.  `print2()` maintains its scroll-back with `pntrLet`/`pntrAddElement`,
so a `bug()` from the pool machinery re-enters `bug()`.  Five of the six
sites cannot continue and now exit.

`bug(1305)` is the one I would ask about.  It stays non-fatal, correctly,
since the next line returns on the same condition, but it now reports with
`printf(BUG_CHECK_FORMAT, 1305)` instead of `bug()`.  That takes the line out
of `print2()`, so it no longer appears in an OPEN LOG session and no longer
goes through the scroll-back, and it can interleave oddly with buffered
output.  It is still greppable, which is what the fuzz harness needs.  Is
losing it from the log acceptable?  It is the only site where a report
survives but changes channel.

**564b19b, tighten the "Not bug()" comments to one shape.**  Accept.
Comments only, no code.

**eb3b2ad, give every bug() site a unique number.**  Accept, verified
mechanically.  Master has 15 numbers used by more than one live site; after
this commit there are none.  The six that still look like duplicates to a
naive grep are one live call plus a commented-out line or a mention in prose
(1348, 1350, 1352, 1522, 2101, 2106).  Each of the eight new numbers is used
exactly once, and each old number is still used by the site that owns the
band.  Only cost: a number printed by an older binary now points somewhere
else, which is worth a line in the release notes.

**8554d26, terminate on fatal conditions instead of continuing or reporting
success.**  Accept.  The four `printLongLine()` traps each index `longLine[]`
out of bounds or leave the enclosing loop unable to end, so returning from
`bug()` was the bug.

Two behavior changes to call out.  mmword.c printed `*** FATAL *** Overflow#n`
and called `exit(0)`, reporting success for a fatal overflow; that becomes
EXIT_FAILURE.  Anything scripted around that exit status sees a change, and
it was wrong before.  It also has a happy side effect: those traps now print
the `?BUG CHECK` line, so `fuzz/run-cases.sh` can see them.  Under the old
text and `exit(0)` the harness would have reported such a case `ok`.

**5a02a14, type errorMessage()'s severity as the enum it is.**  Accept, and
the "no behaviour change" claim holds.  `enum severity {notice_, warning_,
error_, fatal_}` gives `error_` 2 and `fatal_` 3, which are exactly the two
literals the body was comparing against.  Every call site passes an
enumerator.

**0e69d51, minor: streamline initialization.**  Accept.  It replaces a
`static char empty[] = ""` with the literal `""`, which is what `initState()`
should have had; the field is `char const*` and nothing writes through it.
Please give it a body, though.  Every other commit in this PR explains
itself, and "minor" is the one word that stops a reviewer looking.

**897273e, mmfatl.c: detail comment of finishMessage.**  Accept, comment
only.  Same request for a body, or fold it into fefc894, which is the commit
that added the function.

**d415b8a, print the file name, not the build path, in fatal errors.**
Accept.  An out-of-tree build putting the builder's home directory into a
user-facing message is a real complaint, and it eats the 1024 bytes a fatal
message has to fit in.  `#ifdef __FILE_NAME__` is the right test since it is
a predefined macro, and the fallback to `__FILE__` keeps older compilers
working; it arrived in clang 9 and gcc 12, so the fallback is the common case
for a while yet.  Keeping `__FILE__` in mmtest.h is right, that being the
test's own location.

**98aab8f, do not retain the caller's format string.**  Accept.  No UB is
executed, as the message says, but a file-scope static holding a pointer into
a caller's automatic buffer is exactly the kind of thing that becomes UB
after the next edit.  The test is the good part: it overwrites the caller's
buffer, terminator included, which is the only way to tell a retained pointer
from a legitimately NUL-terminated one.

## What a user can see change

All on error or fatal paths.  Nothing on a successful run.

1. `?Error trying to write "(null)"` becomes the real file name.  (4ee5c9b)
2. Fatal conditions in mmvstr, mmdata, mmpars, mminou and mmword now print
   `At file:line`, then the `?BUG CHECK` line, then `*** FATAL ERROR ***`,
   and exit.  Previously most printed the FATAL line first, and some went
   through `bug()`'s full interactive text.  (0f0c069, 3b1a79f, 8554d26)
3. Those sites no longer offer `bug()`'s "I" or "S" escape.  That is the
   point: every converted site was one where continuing was undefined.
4. mmword.c overflow traps exit 1 instead of 0.  (8554d26)
5. `bug(1305)` no longer appears in the session log.  (3b1a79f)
6. Six bug numbers moved, so a number from an older binary now points at a
   different site.  (eb3b2ad)

## Points to raise

- `bug(1305)` losing the log.  See above.  If it matters, it could keep
  `print2()` and drop only the `bug()` call.
- Commit bodies on 0e69d51 and 897273e.
- The stray reformatting of `fSafeOpen()` inside deef140.

## A trap for anyone verifying this

`./build.sh -t` runs `make CFLAGS=-DTEST_ENABLE`, and make will not
recompile objects that are already current.  Run it after a normal build and
you get `metamath_test` relinked from non-test objects: it prints the banner,
runs no tests, and exits 0.  My first run looked like a pass that way, with
both binaries the same 2398072 bytes.

`./build.sh -c -t` is the honest form.  Clean, `metamath_test` is 938368
bytes, prints its 12 tests and passes them.  I confirmed the suite really
bites by breaking one assertion in `test_fatalErrorExitAt`: exit 1 and the
failing assertion named.  Not this PR's doing, but it touches the mmfatl
tests, so anyone checking those needs the `-c`.  CI is safe, since its
regression job builds only with `-t` from a fresh checkout.

## What I did not check

- Only gcc on Linux.  The `PRINTF_FORMAT_ATTR` and `MMFATL_FILE` fallbacks
  are compile-time only and read correctly, but I did not build with clang,
  and not on macOS or Windows.
- I did not read all 21 cast sites in d8acef8 individually, relying on the
  warning-free build for coverage.
- No fuzzing, only the recorded cases.
- Message truncation in mmfatl when a path plus message exceeds 1024 bytes.
  d415b8a makes it less likely rather than impossible.

david-a-wheeler:
So I approve once you've double-checked the concerns listed above.

src/mmfatl.c
benjub: You're right that `__FILE_NAME__` reduces this rather than removes it: it drops the directory, not the basename, so a non-ASCII source file name would still reach the buffer. My "only ASCII" claim holds by convention, not by construction.

I still don't think the logic pays for itself: `mmfatl` never interprets bytes, so there is no UB either way, and the worst case is one glyph at a truncation boundary in an already-fatal message, versus permanent complexity in a module that has to work with a corrupt heap and no locale.

benjub:
> So I approve once you've double-checked the concerns listed above.

I checked, rewrote (with AI assistance) the two missing commit message bodies (hence the force push). As for the other two points:
- `bug(1305)`: I replaced it with `printf(BUG_CHECK_FORMAT, 1305)` and commented why in the code above it: calling `bug(1305)` would cause recursion: `bug` calls `print2` which calls `pntrLet` which calls `addToUsedPool`. Since I use `BUG_CHECK_FORMAT` it stays greppable for `fuzz/run-cases.sh`.
- `fSafeOpen()` : Agreed, it turned out to be an unrelated diff. It's transparent, but increases consistency with other `defined` guards (better not to use #ifdef here, since there is an #elif later, which we cannot change to #elifdef, which is C23).

wlammen:
As far as mmfatl is concerned, the issues in "Commit by commit" seemed to have been addressed.
The trap described in "A trap for anyone verifying this" is addressed in a different pull request.

david-a-wheeler:
Looks good to me!

Some minor notes, none blocking. Consider these for a future PR.

### 1. `PRINTF_FORMAT_ATTR` promises more checking than mmfatl delivers

mmfatl handles only `%s` and `%u`, but the attribute validates against full
printf. Maybe add a line on `PRINTF_FORMAT_ATTR` noting the supported subset?

### 2. `addToUsedPool`'s new `printf` is missing the flush

```c
if (((long *)ptr)[-1] == ((long *)ptr)[-2]) printf(BUG_CHECK_FORMAT, 1305);
```

Every other direct-`printf` bug report in that file follows with
`#if __STDC__ fflush(stdout); #endif`. This one returns rather than exiting,
so the message can be lost. It also bypasses `print2`, so it never reaches the
log file.

david-a-wheeler (approved):
See my comments.

david-a-wheeler:
Let's wait to merge until we've fixed the "metamath website down" problem I inadvertantly caused.

david-a-wheeler:
Also: I suggest merging master *back* into this PR once we merge #198 into master. PR #198 adds some new tests, including downloading and using set.mm and iset.mm (for proof verification HTML generation). That should reduce the chance of another similar problem.

benjub:
> Also: I suggest merging master _back_ into this PR once we merge #198 into master. PR #198 adds some new tests, including downloading and using set.mm and iset.mm (for proof verification HTML generation). That should reduce the chance of another similar problem.

Done, and the tests pass, including the recently added ones from #198.  Merging.

PR #198, opened 2026-08-12
Fix issue 196

### LINKED ISSUE #196: website generation is failing with BUG 1741
When we try to generate the web page for ax-1, metamath-exe fails with a BUG 1741 message.

The offending commit (based on a git bisect run) appears to be https://github.com/metamath/metamath-exe/commit/ae054502ba38569fa546b6e247527642288b00e2

Further reproduction instructions at https://github.com/metamath/metamath-website-scripts/issues/21#issuecomment-5268421800 (and, potentially, the other comments on that issue).

david-a-wheeler:


david-a-wheeler:
@jamesjer @LegionMammal978 @wlammen @benjub - this fixes things by removing the "-1" that should no longer be there. I'd like an approve before merging.

I didn't want to make things *worse*, so I added special tests related to this, and a general set.mm/iset.mm test that verified proofs and generated a few HTML pages. It takes too long to generate *all* HTML pages, but generating a few samples seemed like a reasonable compromise that would systemically prevent similar problems.

david-a-wheeler:
The actual *fix* is tiny and I agree with the original proposal. So if I don't hear in an hour, I'll just merge, since we don't want the website to stay down, and we have multiple people agreeing that this should fix it. I normally wait until someone else reviews, but "website down" seems like a reasonable emergency.

Adding tests doesn't seem like a bad thing, and we can remove them or tweak them later. I just wanted to make sure that I'd tested the fix, not just accepted the fix.

jkingdon (approved):

jkingdon:
Very nice work. I tried to write a test case along the lines of issue196.in but didn't get very far (my initial attempt did not reproduce the problem).

I have downloaded this branch, seen that the tests pass, put back the erroneous ` - 1 `, and seen that both issue-196 and issue-196-silent fail.

david-a-wheeler:
Thanks! Ok, merging. The site will start regenerating soon, so I think it'll just clean up on its own once this is merged.

PR #199, opened 2026-08-14
update file description in mmtest.h, prepare documenting TEST_ENABLE

wlammen:
This PR addresses:

File description:
Generally: reduce duplicated explanations
1. When to use regression tests; automatic runs by GitHub
2. Add a step 5 to restore the usual development environment after executing regression tests.
3. Better explain how an executable with regression tests behaves, and why regression test code is conditionally included.
4. How RUN_TESTS_AND_EXIT_IF_ENABLED behaves when TEST_ENABLE is not /is defined

Prepare a description of TEST_ENABLE. The problem here is that TEST_ENABLE is usually defined outside of the source code, say by option -t of build.sh.  During the generation of documentation, Doxygen encounters the macro TEST_ENABLE and does not know how to interpret it correctly. It always issues a warning, but in addition may exclude test code from documentation altogether. There is a mechanism built in Doxygen that allows setting this macro by its configuration file. Implement this mechanism, and reduce the warning to the current "TEST_ENABLED not documented".

Building process both for the normal and the test executable run fine, running them shows no errors.

The updated documentation looks like this:
grafik

### DIFF
--- a/src/mmtest.h
+++ b/src/mmtest.h
@@ -120,7 +120,7 @@
  * including knowledge about Metamath details in the configuration of an
  * external, non-central tool.
  */
-#define TEST_ENABLE
+  #define TEST_ENABLE
 #endif // DOXYGEN_PARSE
 
 #include <stdio.h>
--- a/src/mmtest.h
+++ b/src/mmtest.h
@@ -46,6 +46,14 @@
  * Failing tests always display diagnostic information identifying the failing
  * assertion and its source location.
  *
+ * 5. After enabling or disabling testing, even without modifying any source
+ * files, all intermediate artifacts must be rebuilt. Pass the \c -c (clean)
+ * option to \c build.sh to enforce a complete recompilation of Metamath:
+ * \verbatim
+   ./build.sh -c
+   ./metamath
+   \endverbatim
+ *
  * \subsection test_details Details and implementation
  * If the macro \c TEST_ENABLE is defined by the \c -t option of \c build.sh,
  * the script compiles the regression tests into an executable named
--- a/src/mmtest.h
+++ b/src/mmtest.h
@@ -84,9 +84,7 @@
  * run automatically by GitHub's checks whenever changes are pushed.
  *
  * The macro \c RUN_TESTS_AND_EXIT_IF_ENABLED should be the first instruction
- * in \c main. It expands to nothing when testing is disabled. When testing
- * is enabled, it runs the regression tests and terminates the program,
- * so the normal program execution does not take place:
+ * in \c main:
  * \code
  * int main(int argc, char *argv[]) {
  *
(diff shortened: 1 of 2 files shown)


wlammen:
I am not sure whether I need to explicitly mark a PR as ready for inclusion. This one is, for 3 days now. Later commits are polishing edits, changing nothing essential anymore.

benjub:
> I am not sure whether I need to explicitly mark a PR as ready for inclusion.

I do not know the policy, that is a question for maintainer @david-a-wheeler.

As I wrote in https://github.com/metamath/metamath-exe/pull/194#issuecomment-5246413967, I have little bandwidth for this project (and limited C skills), and I wrote there that probably larger refactorings which are not bugfixes may not find reviewers' time. Actually, it looks like @david-a-wheeler too has limited bandwidth, which is why I was surprised by his enthousiastic answer in https://github.com/metamath/metamath-exe/pull/188#issuecomment-5187012236. I do not know if other people are willing to give review time.

As for this PR: since it's a follow-up of the one I reviewed a few days ago (#194), I will also review it, but that shouldn't be an indication that I will have time to review other PRs in the future.

wlammen:
In CONTRIBUTING.md the development of the C sources is discouraged.

Here is the reason why I recently updated the documentation: I wondered whether the disjoint variable check covers a newly introduced variable type properly. In the end I got lost in the details, so I started updating the documentation a bit.  That's what these pull requests are about: Mostly documentation, perhaps later patch the portions that cover the disjoint variables, once I have a better understanding, along with test code, of course.

src/mmtest.h
```diff
@@ -33,40 +40,46 @@
...
+ * \verbatim
+   ./build.sh -c
+   ./metamath
+   \endverbatim
+ *
  * \subsection test_details Details and implementation
- * If the macro \c TEST_ENABLE is defined by the \c -t option of \c build.sh,
- * the script compiles the regression tests into an executable named
- * \c metamath_test. The value of \c TEST_ENABLE is irrelevant; only whether
- * the macro is defined matters. Defining the macro directly in this file does
```
benjub, line 46: This was important information in my opinion.

benjub (commented):
Only one comment.

For the rest, I prefer to defer to @david-a-wheeler to judge if that documentation is aligned with the tests/deployment he has recently added.

src/mmtest.h
wlammen: The name of the executable is controlled by the build.sh script, not by the code here, so I removed it here. The general description of invoking the regression tests still provides this information. The rest describes the macro TEST_ENABLE, and is deferred to its description, which would be visible in the next PR.

david-a-wheeler:
> I am not sure whether I need to explicitly mark a PR as ready for inclusion. This one is, for 3 days now. Later commits are polishing edits, changing nothing essential anymore.

We don't have any *special* policy. Typically on GitHub a PR is presumed to be ready *unless* you mark it as draft (on the right-hand-side of the PR click on "Convert to draft"). We also usually wait for others to comment/review.

src/mmtest.h
david-a-wheeler: Okay, with the caveat I'd like to see that next PR :-).

PR #200, opened 2026-08-17
Publish builds (cosmo and wasm)

david-a-wheeler:
Modify the CI system to publish, to a web page, pre-built cosmo and wasm builds.


david-a-wheeler:
The cosmo build is a single build that runs on Windows, MacOS, Linux, and *BSDs for both x86 and ARM.

The wasm build runs on a web browser.

The idea is to add these builds so people can use metamath-exe without being *required* to compile it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,295 @@
...
+        run: env METAMATH=../metamath ./run_test.sh *.in
+
+  unit:
+    name: Unit tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: build
+        run: ./build.sh -t
```
benjub, line 63: Here, you do not need the -c option because you are assured to start with a fresh build directory, correct? If unsure or this can change in the future, then -c does not hurt. If sure, then close/resolve this thread.

david-a-wheeler: Yes, that's correct. Every job runs on its own fresh ubuntu-latest runner with an empty workspace. The `build*/` directory is gitignored so checkout can't bring one in, and nothing caches a build directory (only cosmocc and emsdk are cached). So -c would be a no-op here. I wouldn't expect that to change in this circumstance; it's safer to rebuild each time.

david-a-wheeler:
Any other comments? Approvals? Disapprovals?

tirix:
You have both my thumbs-up, and a heart! as reactions to your earlier comment.
This is a *great* token of support, I believe this PR useful and goes towards the right direction, however I have not taken the time to review the associated code, so I don't feel confident to formally approve it.
I want to try it out once it's live, though.

benjub:
I haven't reviewed and won't have time. What I skimmed through looked good.

david-a-wheeler:
Hearts and no negatives, and there's been a little time, so I think we have enough to merge.

I've been briefly mostly-unavailable, but I'm back, and I think it's time.

Issue #164, opened 2023-09-28
segfaults on 64-bit windows


digama0:
I don't have a windows machine, but the cross compile for 64-bit windows on my machine looks like this:
```
$ cd build
$ make clean
$ make CC=x86_64-w64-mingw32-gcc
$ wine src/metamath.exe
0024:err:file:init_redirects /home/mario/.wine/dosdevices/c:/windows: No such file or directory
002c:err:wineboot:main Cannot set the dir to L"C:\\windows" (2)
002c:err:wineboot:process_run_key Error running cmd L"C:\\windows\\system32\\winemenubuilder.exe -a -r" (2).
wine: could not open working directory L"unix\\home\\mario\\Documents\\metamath\\metamath-exe\\build\\", starting in the Windows directory.
002c:err:shell:SHGetFolderPathAndSubDirW Failed to create directory L"C:\\users\\mario\\Application Data".
002c:err:shell:SHGetFolderPathAndSubDirW Failed to create directory L"C:\\users\\mario".
002c:err:shell:SHGetFolderPathAndSubDirW Failed to create directory L"C:\\users\\mario\\Local Settings\\Application Data".
wine: could not open working directory L"unix\\home\\mario\\Documents\\metamath\\metamath-exe\\build\\", starting in the Windows directory.
*** BUG #1501
Metamath - Version 0.199.pre 29-Jan-2022      wine: Unhandled page fault on read access to 0000000000000000 at address 00000001C8DFD2B0 (thread 0040), starting debugger...
0040:err:seh:start_debugger Couldn't start debugger L"winedbg --auto 60 60" (2)
Read the Wine Developers Guide on how to set up winedbg or another debugger
```
Most of the wine errors are harmless, but metamath immediately gives `BUG #1501` and then dereferences null shortly afterward. The 32-bit windows build works properly.

I believe the reason is because metamath-exe uses `long` as an equivalent to `size_t`, `ptrdiff_t` and the like, but in 64-bit windows `long` is 4 bytes while pointers and allocation sizes are 8 bytes, so there is a lot of invalid casting going around.

creator1creeper1:
Thank you for nothing this issue. Compiling with `clang m*.c -o ../bin/metamath.exe -O3 -funroll-loops -finline-functions -fomit-frame-pointer -m32` (note the `-m32`) works on my machine.

vince-gonzalez:
[...]

PR #203, opened 2026-08-30
Give the null pntrString block the layout its accessors assume

vince-gonzalez:
Part of #164. @tirix asked for a PR, so here is the small one. It is not @wlammen's `pntrInt` change, for the reason on that issue: widening the header means widening `poolFixedMalloc()` and the ~45 other places that read it, and that reaches `nmbrString` too.

**This does not make 64-bit Windows work.** It clears the start-up crash. At least one more LLP64 defect sits behind it, measured at the bottom.

## The defect

`poolFixedMalloc()` writes a block's header as exactly `3 * sizeof(long)` directly below what it returns (mmdata.c:400-407), and `pntrLen()` reads it back by stepping backwards as `(const long *)s` (mmdata.c:2675). `g_PntrNull` imitates such a block for the empty pntrString, so its three header fields have to end where `nullElement` starts.

On LLP64 they do not. `pntrString` is `void*`, so its 8-byte alignment pads the 12-byte header out and puts the element at offset 16. `pntrLen()` reads the padding instead. `print2()` checks that value the first time it runs, which is why the first line the program prints is `BUG #1501`, and it goes on into a null dereference.

Measured by including the project's own headers, so these are the real structs and the real accessors, not a re-typing of them:

```
                                      x86_64-windows   x86-windows
  sizeof(long) / sizeof(void*)              4 / 8          4 / 4
  header written by poolFixedMalloc          12 B           12 B
  offsetof(nullPntrStruct, nullElement)        16             12
  offsetof(nullNmbrStruct, nullElement)        12             12
  pntrLen(NULL_PNTRSTRING)      must be 0      -1              0
  nmbrLen(NULL_NMBRSTRING)      must be 0       0              0
```

`g_NmbrNull` is fine because a `nmbrString` is a `long`, so nothing is padded. Only the pntrString block breaks, which is why the bug is specific to pntrStrings on Windows.

## Why packing

The accessor needs the element at offset 12 and `alignof(void*)` is 8. 12 is not a multiple of 8, so for this struct those two cannot both hold, and packing is what gives up the alignment rather than the offset.

Giving it up costs nothing here, because the allocator already has. `poolFixedMalloc()` returns `malloc()`'s result advanced by 12 bytes, so on LLP64 every pooled pntrString in the program is off a `void*` boundary by 4. Calling the real allocators from a probe:

```
  &g_PntrNull.nullElement % 8   before   0     <- the odd one out
  poolFixedMalloc() data % 8             4
  poolMalloc()      data % 8             4
  &g_PntrNull.nullElement % 8   after    4
```

The static block now matches the allocated ones instead of differing from them.

`NULL_PNTRSTRING` is reached by offset rather than with `&`, because taking the address of a packed member reports every use of the macro as an unaligned pointer — 124 of them under the `-Wall -Wextra` in configure.ac:68. By offset there are none.

`PACKED_ATTR` follows the `#if defined(__GNUC__)` pattern in mmfatl.h, but unlike those attributes it is load-bearing, so it is paired with `_Static_assert` on the offsets: a compiler that ignores it fails the build with a message rather than shipping a broken binary. CompCert is unaffected either way — it targets ILP32 and LP64, where the assert holds with no packing at all. The asserts use the same `__STDC_VERSION__ >= 201112L` guard as the existing one at mmdata.c:34.

Two smaller things in the diff. `g_PntrNull`'s size fields were `sizeof(long)`; they mean "the size of the one element this block holds", which is a `pntrString`. That is a correction of intent, not of behaviour — the old spelling also yields 0 on LLP64, by truncating `-4/8`. And the `\bug` note on `nullPntrStruct` described this exact problem, so it now describes the resolution.

## Results

`zig cc` for `x86_64-windows-gnu` and `x86-windows-gnu`, `-Wall -Wextra`, run on Windows 11.

```
                       baseline 64-bit        this branch 64-bit
  start-up             BUG #1501, SIGSEGV     reaches the prompt
  read set.mm          never gets there       51,140,933 bytes, no errors
  verify proof *       never gets there       runs to 100%, then 0xC0000005
```

`tests/run_test.sh *.in`, comparing `.produced` against `.expected` with line endings normalised:

```
                          pass   fail   no output
  baseline  32-bit          39      2           0
  this PR   32-bit          39      2           0
  baseline  64-bit           0      0          41
  this PR   64-bit          33      0           8
```

32-bit is untouched, including the same two pre-existing failures (`issue196`, `issue196-silent`) either side. 64-bit goes from every test dying at start-up to 33 passing and none failing. `metamath_test` passes on both targets. Warning count is unchanged: the same 3 pre-existing ones before and after.

## What is still broken

The 8 tests that still produce nothing on 64-bit, and the `0xC0000005` at the end of a full `verify proof *`, are a different defect. It is not this change: a variant with no packing anywhere — `unsigned char nullElement[sizeof(pntrString)]`, alignment 1, which lands at offset 12 on its own — crashes in exactly the same place. It was unreachable before only because the program died at start-up. `iset.mm`, `nf.mm` and `ql.mm` verify clean on 64-bit; only the full set.mm run reaches it. I have not tracked it down, so #164 should stay open for it.

### DIFF
--- a/src/mmdata.c
+++ b/src/mmdata.c
@@ -12,6 +12,7 @@
  * to BASIC string functions; memory management; converts between proof formats
 */
 #include <stdarg.h>
+#include <stddef.h>
 #include <stdlib.h>
 #include <string.h>
 #include "mmvstr.h"
@@ -34,6 +35,24 @@
 _Static_assert(sizeof(struct statement_struct) == 256,
     "statement_struct grew; see hasVarWithoutHyp in mmdata.h");
 #endif
+#endif
+
+/* g_NmbrNull and g_PntrNull stand in for a block from poolFixedMalloc(), which
+   writes its three administrative values into the 3 * sizeof(long) directly
+   below the data.  nmbrLen(), nmbrAllocLen(), pntrLen() and pntrAllocLen()
+   read them back from there, so the element has to sit at exactly that offset
+   in these structs too.  Unlike the size check above this is a correctness
+   property and holds on every platform: where it failed -- LLP64, where a
+   naturally aligned void* is padded out to offset 16 -- pntrLen() returned -1
+   for the empty string and metamath died during start-up (issue #164).  */
+#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
+_Static_assert(offsetof(struct nullNmbrStruct, nullElement)
+        == 3 * sizeof(long),
+    "g_NmbrNull's header must be the 3 longs poolFixedMalloc() writes");
+_Static_assert(offsetof(struct nullPntrStruct, nullElement)
+        == 3 * sizeof(long),
+    "g_PntrNull's header must be the 3 longs poolFixedMalloc() writes"
+    " -- see PACKED_ATTR in mmdata.h");
 #endif
 
 /*E*/long db=0,db0=0,db2=0,db3=0,db4=0,db5=0,db6=0,db7=0,db8=0,db9=0;
@@ -77,7 +96,12 @@ long g_sourceLen;
 struct nullNmbrStruct g_NmbrNull = {-1, sizeof(long), sizeof(long), -1};
 
 // Null pntrString
-struct nullPntrStruct g_PntrNull = {-1, sizeof(long), sizeof(long), NULL};
+// The two size fields are the size of the one element the block holds, which
+// is a pntrString and not a long; they differ on LLP64.  Both spellings happen
+// to give pntrLen() == 0 there, the old one by truncating -4/8, so this is a
+// correction of intent rather than of behaviour.
+struct nullPntrStruct g_PntrNull
+    = {-1, sizeof(pntrString), sizeof(pntrString), NULL};
 
 // nmbrString memory allocation/deallocation
 temp_nmbrString *nmbrTempAlloc(long size);
--- a/src/mmdata.h
+++ b/src/mmdata.h
@@ -12,6 +12,8 @@
  * \brief includes for some principal data structures and data-string handling
  */
 
+#include <stddef.h> // offsetof, used by NULL_PNTRSTRING
+
 #include "mmvstr.h"
 
 // debugging flags & variables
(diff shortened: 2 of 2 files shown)


tirix:
Thanks for the PR!
I don't have the rights to launch the workflow, maybe @david-a-wheeler can?

david-a-wheeler:
I can, and I just did!

src/mmdata.c
```diff
@@ -36,6 +37,24 @@ _Static_assert(sizeof(struct statement_struct) == 256,
...
 #endif
 
+/* g_NmbrNull and g_PntrNull stand in for a block from poolFixedMalloc(), which
+   writes its three administrative values into the 3 * sizeof(long) directly
+   below the data.  nmbrLen(), nmbrAllocLen(), pntrLen() and pntrAllocLen()
+   read them back from there, so the element has to sit at exactly that offset
+   in these structs too.  Unlike the size check above this is a correctness
+   property and holds on every platform: where it failed -- LLP64, where a
+   naturally aligned void* is padded out to offset 16 -- pntrLen() returned -1
+   for the empty string and metamath died during start-up (issue #164).  */
```
tirix, line 47: I guess by "administrative values" you mean that the structures are meant to provide some replacement values for the values which would have been provided by `poolFixedMalloc`, but I think the term "administrative" is not quite the right one here.

Maybe someone who better masters english can help here?

tirix (approved):
Overall I think this is the right change, and you tell us it works (I could not test as I also don't have a PC), so this is good for me.

I believe the comments could be shorter and more to the point, but that might just be my taste.

vince-gonzalez:
Thanks for the review.

On the wording: "administrative" is the file's word rather than mine. mmdata.h uses it ten times, twice inside this struct's own doc block ("the hidden administrative values of such block"), so I matched it. If it should change, changing it only in my lines would leave them disagreeing with the neighbours that use it. Your call on the word, and I will use whatever you pick.

On length you are right. The `nullPntrStruct` block is 14 lines and can be about half that without losing the reasoning. Say how short you want it and I will cut it and apply the wording in one push.

src/mmdata.c
tirix: > On the wording: "administrative" is the file's word rather than mine. mmdata.h uses it ten times, twice inside this struct's own doc block ("the hidden administrative values of such block"), so I matched it. 

Ok, thanks, so I misunderstood "administrative". The variables are administrative, rather than the value, in the sense that they keep the record for how many bytes are allocated, actually used, etc.

david-a-wheeler:
For the moment, using the file's internal terminology makes sense to me. They aren't what I would have chosen, but if we change them we need to make sure we change consistently.

david-a-wheeler:
Thanks so much! I suspect we'll end up merging this, but I want to think a little about this first.

For now people on Windows OS 64-bit can just use the precompiled metamath.exe since it uses the LP64 model (long==pointer==8 bytes). This commit is a small step towards it also supporting the LLP64 model (used by "Windows 64").

I'm disappointed that Norm presumed long==pointer. I come from a world where computers were more different (e.g., bytes weren't always 8 bits, words weren't always a power of 2, void * and long* aren't always the same, etc.), so I often presume that types are different unless they are *textually* identical (which is overkill, but has also saved my butt). I think we can fix that long term, and we should.

